PR #19, opened 2019-07-22
Python 2 compatibility

w-k-jones:
Small change to formatting allows python 2 compatibility. Example notebooks only work in python 3 due to changes to urllib


mheikenfeld:
That's great, I was not aware that it only required such small changes to make things work in python 2. If you find a good way to make the example notebooks work in Python 2, go ahead, we could also add a hint to download the data manually when using Python 2 for now...

w-k-jones:
I think the example notebooks can be made compatible using the six module, I'll have a look into it. Downloading the data was the only problem

Cheers, Will

On 24 Jul 2019, at 15:59, Max Heikenfeld <notifications@github.com<mailto:notifications@github.com>> wrote:


That's great, I was not aware that it only required such small changes to make things work in python 2. If you find a good way to make the example notebooks work in Python 2, go ahead, we could also add a hint to download the data manually when using Python 2 for now...

PR #20, opened 2019-07-24
Python 2/3 compatibilty fixes for example notebooks

w-k-jones:
Tested with both python 3.7 and 2.7 installs


w-k-jones:
The animation files are not present as they don't generate correctly for me on jasmin. Might be worth re-running them to get them back

PR #27, opened 2019-09-10
Rewrite feature detection

mheikenfeld:
I have restructured and improved multithreshold feature detection:
-  Split up function into separate subfunctions for each timestep and threshold value as well as for some underlying calculations such as the checks regarding features from different threshold values
- Removed performance bottlenecks related to unnecessary numerous creation of pandas dataframes or too many individual assignments of values into dataframes and looping over combinations of features for the removal of features superseded by higher-threshold features 
- Added some additional tests making sure that feature detection for the sample data returns a dataframe

Result: 
- massive speedup of feature detection step by at least an order of magnitude
- no change in the results, since there were no changes to the actual calculations


mheikenfeld:
I have noticed that I had accidentally merged some of the changes to the segmentation into the branch I used for this development, I have removed that now, so that this pull request is a clean thing only including the changes to the feature detection.

PR #38, opened 2020-06-04
Fix docstrings and adjust configuartions

w-herbst:
The documentation should now be parsed properly by Read the Docs:
https://w-herbst-tobac.readthedocs.io/en/docs/modules.html


w-herbst:
That was on purpose. The routine listings were not parsed, but since we are using autosummary we have routine listings anyway. I only added them in the two files to test the functionality.

PR #41, opened 2020-12-11
Integrating TINT into tobac

rcjackson:
Here is the new PR to the 2.0-dev branch for placing TINT into tobac.


w-k-jones:
Just going through and checking everything is running correctly now. Found that the travis-ci build is failing because pyproj isn't listed as a conda dependency, I've fixed this in the main repository but it might be good to fix that in the pull request as well to see if the CI runs fine otherwise

w-k-jones:
Got everything running fine, and it seems to work correctly. However, it looks like tint.make_tracks is missing the last scan in the dataset, is this by design?

PR #46, opened 2021-02-06
Speed up linking_trackpy

zxdawn:
`linking_trackpy` uses too many loops which will cost more time for large datasets.

Detailed tests are listed in the comments ...


zxdawn:
The length of the `Track` example data is 37754.

## Old method

```
1min 34s ± 228 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

## New method

```
1.54 s ± 523 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

PR #51, opened 2021-05-20
Support setting interpolation kind in `add_coordinates`

zxdawn:
1. Support setting interpolation kind of coord
2. Use `xarray.DataArray` instead of `iris.Cube`:
    - Support ms in the time variable

### DIFF
--- a/tobac/themes/tobac_v1/feature_detection.py
+++ b/tobac/themes/tobac_v1/feature_detection.py
@@ -27,6 +27,7 @@ def feature_detection_multithreshold(field_in,
                                      min_num=0,
                                      target='maximum',
                                      position_threshold='center',
+                                     coord_interp_kind='linear',
                                      sigma_threshold=0.5,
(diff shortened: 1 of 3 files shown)


zxdawn:
Note that the maximum resolution of `iris.unit.num2date()` is 1 second.
So, we should use `round()` to convert the time.

However, there's a special issue:
`2020-06-09 00:01:14.500` is converted to `2020-06-09 00:01:15` by num2date, which is different from `round()` (see the [pandas issue](https://github.com/pandas-dev/pandas/issues/41796)).
So, we need to check the ms manually:
```
        round_time = features['time'].dt.floor("S") +\
                     pd.to_timedelta((features['time'].dt.microsecond>=5e5).astype('int'), 'S')
```

zxdawn:
I close this PR and will split it into different PRs because it includes different functions: adding interpolation kind and switching to partly xarray.

PR #55, opened 2021-07-14
FIX: Wrong id being placed in cell_mask.

rcjackson:
The intermediate cell id (pre-matching) was being placed into cell_mask instead of the proper UID. This is fixed now.


fsenf:
Hi @rcjackson, thanks for your fix! 

I would take responsibility for your PR. Currently, I struggle to understand your commit history. I guess you accidentally started your fix with the master branch? 

Cheers, Fabian.

rcjackson:
I did at first, but then changed. Though I wonder if I need to take the tracks.py file out of this PR, close this PR, start a new branch from v2.0-dev then commit tracks.py to v2.0-dev to make the commit history more clear.

fsenf:
Bobby, it is good as it is. If I understand it right, our master also went forward (just documentation related stuff) and this update is also included in your PR. We need to get this in our v2.0-dev branch anyway. So, we can do it at once with your PR

tobac/examples/eric_test.py
```diff
@@ -0,0 +1,9 @@
+from tobac.themes import tint
+from copy import deepcopy
+track_params = deepcopy(tint.objects.default_params)
+
+nc_file_path = ('/home/rjackson/tracer-jcss/Tgrid_*.nc')
```
fsenf, line 5: it would be great if others could use the data and repeat the tests ... is there an option to share the data?

tobac/themes/tint/tracks.py
```diff
@@ -72,8 +72,12 @@ def make_tracks(grid_ds, field, params=None):
         record.add_uids(current_objects)
         tracks = write_tracks(
             tracks, record, current_objects, obj_props)
-        cell_mask[i] = frame1
-         
+        uids = np.array([int(x) for x in current_objects['uid']])
```
fsenf, line 75: this all looks good, but without testing capabilities it is hard to tell more ;-)

fsenf (commented):
Bobby, 

it all looks good! I would merge it, but before 

* I would like to get rid of the *pyc files in the tint/__pycache__ directory
* Could you remove them and update your PR?

Thanks, Fabian

PS: Is there an option to share testing data (later)? Could we chat on this topic?

fsenf (approved):
OK, I will let the PR go and remove __pycache__ files thereafter. I did not find that the other changes break the functionality ... However, we need more test on this ...

PR #56, opened 2021-09-08
Add count of neighboring cells

jcssouza:
This adds the number of neighboring objects to the the function check_isolation instead of just returning true or false for a cell. The addition of this metric in the final xrarray output was suggested by Jiaxi Hu and @deeplycloudy.


w-k-jones:
Just updated this to merging into v2.0-dev branch, which seems to have cleared up all the conflicts. I'll review and merge once I get the chance, but it all looks fine at first glance

tobac/themes/tint/objects.py
```diff
@@ -31,17 +31,20 @@ def check_isolation(raw, filtered, grid_size, params):
                                       params['ISO_THRESH'])
     nobj_iso = np.max(iso_filtered)
     iso = np.empty(nobj, dtype='bool')
+    neighbors = np.ones(nobj)*(-9999)
```
freemansw1, line 34: Why is this set to -9999? I'm not that familiar with the tint code yet, so if there's something obvious, let me know. Trying to figure out missing data vs. 0 neighbors; is this better as 0?

JuliaKukulies:
Converted this to a draft, as this change is outdated with our current development

PR #60, opened 2022-01-14
Updated Testing and CI to use GitHub Actions and Docker

freemansw1:
As we discussed today, these are the changes that we have made to allow CI through GitHub Actions and Docker for our testing. I don't think this is completely ready to be merged (even into the new `dev` branch that I just created), but I mostly opened this to start the discussion and work on this. 

Note that this only checks the newest version of Python/Libraries, so some thoughts:

- Are we still supporting Python 2.x? I don't see a reason to personally, but thought it would be worth discussing. 
- What other versions of libraries should we test?
- What is the best way of testing multiple libraries in this framework? I'm not a Docker expert (or even intermediate), so I would love input here. 

Note that when merging, I recommend doing a squash commit because my bookkeeping was bad on my fork.


w-k-jones (approved):

w-k-jones:
Have merged the PR to get things going for now. Regarding questions:

1. No reason to keep python 2 support, we should drop it. I think I was the one who used it in the first place for running some old code, but have no reason to now
2. I think it's best to focus on the most recent versions of dependencies, so that we don't get stuck with depreciated code. If people have a particular reason to use an older version of a dependency that breaks things it's probably easier to fix that individually than try and keep backward-compatibility for everything

PR #66, opened 2022-02-01
Changed call to calculate_area to fix a bug

### LINKED ISSUE #50: area_histogram
https://github.com/climate-processes/tobac/blob/730fa9affc0b6e7d3facf525faf73cc3a6406e74/tobac/analysis.py#L319

should be:

features=calculate_area(features,mask,method_area)

freemansw1:
This should resolve #50. Ultimately, it would be good to add more tests for the analysis code to avoid these issues.


fsenf (approved):

fsenf:
OK, this seems to be easy to solve for me ;-)

Issue #62, opened 2022-01-18
Cell IDs are floats, while feature IDs are ints


deeplycloudy:
In the `tobac_v1` theme's tracking step, a cell ID is assigned, and the convention is to [use NaN to mark that a feature has not been assigned to a cell track](https://github.com/climate-processes/tobac/blob/v2.0-dev/tobac/themes/tobac_v1/tracking.py#L228), so the cell ID for each feature has to be a float. 

However, for the segmentation mask the convention is to [use 0 if no feature is found at a grid box](https://github.com/climate-processes/tobac/blob/v2.0-dev/tobac/themes/tobac_v1/mask_segmentation.py#L83), so the feature ID can be an integer.

I'd prefer integer IDs, since those aren't subject to float rounding problems in the case of very large integers, and they compress better. The downside is having to adopt a particular integer to mark missing data, though there are [CF conventions](https://cfconventions.org/cf-conventions/cf-conventions.html#missing-data) for that, i.e., the `valid_range` and `_FillValue` attributes.

Would the maintainers be open to switching the cell IDs to integers?

freemansw1:
We've had the same discussions here at CSU. My thought is to have this be a user-set parameter (perhaps defaulted to `np.nan` to keep compatibility?), but I'm not sure what everyone else thinks.

freemansw1:
Worth noting that if we switch this over to a user-set parameter, it would be a good idea to do the same for segmentation.

deeplycloudy:
One more clarification: tracking output is actually an object array like this: `array([nan, 2, 3], dtype=object)` I thought it had a float dtype, but that was an artifact of having saved the tracking data using `.to_netcdf`, which seems to have cast it to float.

zxdawn:
@deeplycloudy Yes, Because np.NaN is float, it forces the entire array of integers to become floating-point numbers.
https://github.com/pydata/xarray/issues/6091#issuecomment-998789248

galexsky:
[...]

PR #74, opened 2022-02-18
Added the ability to set a custom stub cell value instead of `np.nan`

freemansw1:
This should resolve (most of) #62. Rather than set the stub/non-tracked cells as `np.nan`, `tobac.tracking.linking_trackpy` now accepts `stub_set_number` as a parameter for the value of `cell` to set the stub/non-tracked cells to, with a default value of -1.


deeplycloudy:
I tested this (after porting to v2.0-dev), and it resolves the float-coercion problem in #62.

By definition, the lists of segmentation IDs and feature IDs  are separate, so I agree segmentation is a separate discussion.

Is "stub" a term in the public API, or a term internal to the implementation? For symmetry with other kwargs, I think I'd prefer something like `cell_number_unassigned` instead of `stub_set_number`. That mirrors `cell_number_start`.

deeplycloudy:
Also, thinking about metadata: it would be good to capture the user's chosen stub/unassigned number as an attribute. I could propose how to do that for xarray, but not sure if iris supports such metadata.

I'm very interested in having the metadata. When working with a combined dataset (features, the feature mask, and cells) one would need to know which ID to ignore, and this function seems like the right place to add that metadata so it can be used appropriately downstream. While the user knows the ID when creating the dataset, that information is lost when those data are saved to an archive apart from the software.

If we think about shipping those data to an archive,

freemansw1:
"stub" is indeed used in the public API (see `tobac.tracking.linking_trackpy`), although it definitely shouldn't be as the language isn't intuitive. I'll switch to `cell_number_unassigned` as you suggest here. 

The issue with metadata here is actually with pandas rather than iris, as in v1.x, the feature/tracking data comes out as a pandas DataFrame. Pandas has the `DataFrame.attrs` dict, but it is described as experimental. For now, I'm going to add the selected unassigned cell value there as `cell_number_unassigned`. 

I (or someone else) will need to port this to v2.x, but I'm hesitant to make any ports until we get to v1.5.

freemansw1:
I've moved this to the v1.3 milestone to match the issue (#62).

freemansw1:
Per our discussion last week, waiting on #78 to be merged before merging this.

freemansw1:
Now that #89 has been merged, I need to update this PR to avoid the conflicts.

freemansw1:
I've updated it to avoid the conflict and I will re-request a review to keep with good practices before merging.

w-k-jones (approved):

PR #76, opened 2022-02-18
Adds isolation in 2d to segmentation file.

kelcyno:
This is just an addition to the main tobac segmentation file. The additional parameter is ISO_dilate, a value by which the 2d feature is dilated in order to find the number of neighbors. 1 Neighbor value indicates the feature (which was dilated) does not have a neighbor. Values greater than 1 are the number of neighbors in addition to the base feature. Non-positive values indicate a feature area was not created for the feature id and thus no valid neighbors.


freemansw1:
It looks like the CI failed due to switching between spacing and tabs for indents. I believe that the rest of the code uses four spaces to denote indentation rather than tabs. Looking at your code, I suspect that that may be the only issue?

tobac/segmentation.py
```diff
@@ -25,13 +25,15 @@ def segmentation_timestep(field_in,features_in,dxy,threshold=3e-3,target='maximu
     method:     string
                 flag determining the algorithm to use (currently watershedding implemented)
     max_distance: float
-                  maximum distance from a marker allowed to be classified as belonging to that cell
+                maximum distance from a marker allowed to be classified as belonging to that cell
+    ISO_dilate: int
```
freemansw1, line 29: Why is 8 the default here? I'm not opposed to the value, I'm just curious.

tobac/segmentation.py
```diff
@@ -120,6 +123,33 @@ def segmentation_timestep(field_in,features_in,dxy,threshold=3e-3,target='maximu
     else:                
         raise ValueError('unknown method, must be watershed')
 
+	#create isolation, currently only available for 2d tracking
+    if field_in.ndim==2:
```
freemansw1, line 127: Do we want to make this entirely optional? Is running the isolation here necessary for everyone, or would it be better to have a parameter in place to turn it on/off?

freemansw1: I guess the question here ultimately is: what's the performance penalty?

freemansw1:
I've added some comments to the code, but generally I'm happy here. I think we will want to wait to merge this until after #78 , but once that's in and this is updated to match the new style, I'm happy to merge this.

w-k-jones:
Overall this looks good, but I have a couple of thoughts/requests:

1. Like Sean, I think it would be better to have this as an optional parameter to calculate, rather than being run for everyone. Perhaps it would be good to set the default value of "ISO_dilate" to 0 and only calculate the isolation metric if a value greater than 0 is provided?
2. Could this be run separately from segmentation? I.e. could this be run as a separate analysis function on a previously calculated segmentation mask. This would negate the need for the first point, and allow the effects of different ISO_dilate values to be compared more easily

The merge conflict is just due to formatting, so we can fix that easily before merging.

freemansw1:
I've marked this PR as a draft for now.

PR #81, opened 2022-03-10
New testing utilities and feature detection tests

freemansw1:
As part of our efforts here on speedups and 3D/PBC changes, we've also started to beef up the test cases. This PR contains several new utility functions for testing (that are generally, themselves, tested) and a simple test case for feature detection. We're planning on continuing to expand on this, producing similar tests for tracking and segmentation. For tracking, I'm waiting until #74 is resolved as there are some headaches dealing with the changing data types.

### DIFF
--- a/tobac/testing.py
+++ b/tobac/testing.py
@@ -405,8 +405,8 @@ def make_feature_blob(in_arr, h1_loc, h2_loc, v_loc = None,
         start_loc = 1
         v_min = 0
         v_max = in_arr.shape[start_loc]
-        start_v = round(max(v_min, v_loc - v_size/2))
-        end_v = round(min(v_max-1, v_loc + v_size/2))
+        start_v = int(np.ceil(max(v_min, v_loc - v_size/2)))
+        end_v = int(np.ceil(min(v_max-1, v_loc + v_size/2)))
         if v_size > v_max - v_min:
             raise ValueError("v_size larger than domain size")
 
@@ -423,11 +423,11 @@ def make_feature_blob(in_arr, h1_loc, h2_loc, v_loc = None,
         raise ValueError("Horizontal size larger than domain size")
 
     # let's get start/end x/y/z
-    start_h1 = round(h1_loc - h1_size/2)
-    end_h1 = round(h1_loc + h1_size/2)
+    start_h1 = int(np.ceil(h1_loc - h1_size/2))
+    end_h1 = int(np.ceil(h1_loc + h1_size/2))
 
-    start_h2 = round(h2_loc - h2_size/2)
-    end_h2 = round(h2_loc + h2_size/2)
+    start_h2 = int(np.ceil(h2_loc - h2_size/2))
+    end_h2 = int(np.ceil(h2_loc + h2_size/2))
     in_arr = set_arr_2D_3D(in_arr, amplitude, start_h1, end_h1, 
                            start_h2, end_h2, start_v, end_v)
     return in_arr    
--- a/tobac/testing.py
+++ b/tobac/testing.py
@@ -3,63 +3,85 @@ import numpy as np
 from xarray import DataArray
 import pandas as pd
 
-def make_simple_sample_data_2D(data_type='iris'):
+
+def make_simple_sample_data_2D(data_type="iris"):
     """
-    function creating a simple dataset to use in tests for tobac. 
+    function creating a simple dataset to use in tests for tobac.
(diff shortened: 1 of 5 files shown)


freemansw1:
Note that I'm fine updating once #78 is merged, but I wanted to get this in before the development meeting tomorrow.

freemansw1:
I decided to add back in the tracking tests, but forcing the `cell` column to float instead of object.

tobac/tests/test_feature_detection.py
```diff
@@ -0,0 +1,29 @@
...
+    test_data = np.zeros((50,50))
+    test_data = testing.make_feature_blob(test_data, 20, 20, h1_size=5, h2_size=5, amplitude=2)
+    test_data_iris = testing.make_dataset_from_arr(test_data, data_type='iris')
+    fd_output = feature_detection.feature_detection_multithreshold_timestep(test_data_iris, 0, 
+        threshold=[1.5,], min_num=2)
+
+    # Make sure we have only one feature
+    assert len(fd_output.index) == 1
+    # Make sure that the location of the feature is correct
+    assert fd_output.iloc[0]['hdim_1'] == 19.5
```
w-k-jones, line 26: As these are floats it might be worth changing the equivalence statements to using np.isclose instead.

tobac/testing.py
```diff
@@ -327,3 +328,218 @@ def make_sample_data_3D_3blobs(data_type='iris',invert_xy=False):
         sample_data=DataArray.from_iris(sample_data)
     
     return sample_data
+
+def make_dataset_from_arr(in_arr, data_type = 'xarray'):
```
w-k-jones, line 332: It may be advantageous to use the pytest feature to make test dataset functions into decorators to easily allow them to be used across many test functions. Happy with this for now though

tobac/tests/test_feature_detection.py
```diff
@@ -0,0 +1,29 @@
...
+    Tests ```tobac.feature_detection.feature_detection_multithreshold_timestep
+    '''
+    import numpy as np
+    from tobac import testing
+    from tobac import feature_detection 
+    # start by building a simple dataset with a single feature and seeing 
+    # if we identify it
+    
+    test_data = np.zeros((50,50))
+    test_data = testing.make_feature_blob(test_data, 20, 20, h1_size=5, h2_size=5, amplitude=2)
```
w-k-jones, line 18: Might be good to set the feature location as a variable to make it clearer, and link the test value of the assertion to this. Also is it possible to set it up so that the calculated feature locations is the same as the test data location to make this clearer?

tobac/testing.py
freemansw1: I generally agree with this. I need to dig into that feature a bit more.

tobac/tests/test_feature_detection.py
freemansw1: Yeah, this is a good point and has caused a headache. I've switched it from rounding the `start_h1` etc. to consistently running `np.ceil` in `tobac.testing.make_feature_blob` to resolve this.

tobac/tests/test_feature_detection.py
freemansw1: I've switched to using `pytest.approx` here, good catch!

freemansw1:
I've gone ahead and run black on the three new files (`tobac/tests/test_feature_detection.py`, `tobac/tests/test_tracking.py`, and `tobac/tests/test_testing,py`

tobac/testing.py
```diff
@@ -405,81 +509,92 @@ def make_feature_blob(in_arr, h1_loc, h2_loc, v_loc = None,
...
-def set_arr_2D_3D(in_arr, value, start_h1, end_h1, start_h2, end_h2,
-                  start_v = None, end_v = None):
-    '''Function to set part of `in_arr` for either 2D or 3D points to `value`.
-    If `start_v` and `end_v` are not none, we assume that the array is 3D. If they 
-    are none, we will set the array as if it is a 2D array. 
+    start_h1 = int(np.ceil(h1_loc - h1_size / 2))
+    end_h1 = int(np.ceil(h1_loc + h1_size / 2))
+
+    start_h2 = int(np.ceil(h2_loc - h2_size / 2))
+    end_h2 = int(np.ceil(h2_loc + h2_size / 2))
```
w-k-jones, line 533: Happy with these changes, this should also make the sizes more consistent than those given by using round

w-k-jones (approved):

PR #85, opened 2022-03-16
Bug fix for spectral filtering feature

JuliaKukulies:
Bug fix for ```spectral_filtering``` in ```utils/general.py```  that belongs to PR #65

### DIFF
--- a/tobac/analysis/__init__.py
+++ b/tobac/analysis/__init__.py
@@ -1,7 +1,10 @@
-from .analysis import cell_statistics_all,cell_statistics
+from .analysis import cell_statistics_all, cell_statistics
 from .analysis import lifetime_histogram
 from .analysis import calculate_velocity, velocity_histogram
 from .analysis import histogram_featurewise, histogram_cellwise
 from .analysis import calculate_overlap
 from .analysis import calculate_area, area_histogram
-from .analysis import calculate_nearestneighbordistance, nearestneighbordistance_histogram
+from .analysis import (
+    calculate_nearestneighbordistance,
+    nearestneighbordistance_histogram,
+)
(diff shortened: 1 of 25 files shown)


w-k-jones (commented):
Happy to approve this, but it should probably be made clear that this is also formatting all of the v2.0-dev branch. I have no problem with this, but it's probably best to make anyone else working on v2.0-dev aware of this!

JuliaKukulies:
Wonderful, thanks! And apologies that the formatting came in the same commit. I realized that we have to be a bit careful with the formatting so the review does not get too messy. I hope you saw that there were only minor code changes for ```spectral_filtering``` in ```utils/general.py```. 

Ok, agreed! @fsenf @freemansw1 was there any special reason why you wanted the formatting only in ```dev``` first?

freemansw1:
I'm happy to have the formatting changes in `v2.0-dev`; I think my initial thought was that we should wait until 1.5 is feature-complete, but I don't think that's necessary.

w-k-jones (approved):

w-k-jones:
Sorry for the delay catching up with this. I think reformatting v2.0-dev is better sooner rather than later, so will merge this unless anyone has any objections

freemansw1:
Is there a way to have the tests re-run now that we've fixed them not running on this branch?

JuliaKukulies:
Do you mean how to manually run the workflows from the other branches on ```v2.0-dev``` ?  I dont think it is possible, when we do not have any workflow files here.

freemansw1:
Good point! It looks like I put the dockerfile in `v2.0-dev`, but not the GitHub Actions script. I cloned your PR and ran the Dockerfile manually, and the tests passed. I'm happy to merge this now, but will let @w-k-jones as it was his review.

PR #89, opened 2022-03-18
Added tracking speedups

freemansw1:
As discussed a few times in different presentations, we have sped up tracking substantially (and made it scale linearly). A simple comparison, tracking on a single day of GOES-16 IR data on a modern server processor can be seen here:
tracking_times_only. I should note that I think these changes are similar to those done in #46, although it looks like we took slightly different approaches. I haven't benchmarked my changes vs. those in #46, but I can if we think it's necessary. 

Essentially, the theme of the changes was to remove any nested loops, reducing our post-`trackpy` time complexity to O(N) rather than the O(N^2) that it was before. Where loops are necessary, I've tried to use `np.vectorize` to at least allow the loops to be vectorized. 

Longer-term, I'd like to do another PR to clean up some of the older code in this module. For example, `fill_gaps` doesn't
seem to be used in any tobac version, and is a good candidate for removal.

### DIFF
--- a/tobac/tracking.py
+++ b/tobac/tracking.py
@@ -121,7 +121,10 @@ def linking_trackpy(
     ):
         cell = int(i_particle + cell_number_start)
         particle_num_to_cell_num[particle] = int(cell)
-    remap_particle_to_cell_vec = np.vectorize(remap_particle_to_cell_nv)
+
+    remap_particle_to_cell_vec = np.vectorize(
+        lambda particle_cell_map, input_particle: particle_cell_map[input_particle]
+    )
     trajectories_unfiltered["cell"] = remap_particle_to_cell_vec(
         particle_num_to_cell_num, trajectories_unfiltered["particle"]
     )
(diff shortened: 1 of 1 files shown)


tobac/tracking.py
```diff
@@ -239,21 +240,36 @@ def fill_gaps(
...
-        for i, row in track.iterrows():
-            t.loc[i, "time_cell"] = row["time"] - track_0.loc[track_0.index[0], "time"]
-    # turn series into pandas timedelta DataSeries
+
+    t["time_cell"] = t["time"] - t.groupby("cell")["time"].transform("min")
     t["time_cell"] = pd.to_timedelta(t["time_cell"])
     return t
+
+
+def remap_particle_to_cell_nv(particle_cell_map, input_particle):
```
w-k-jones, line 263: Would it be better to make this as a lambda function in or just before line 124 for clarity?

freemansw1: Good point. I've switched this to a lambda function now.

w-k-jones (approved):

PR #99, opened 2022-03-27
Remove `is` comparisons with a string literal

### LINKED ISSUE #97: Removing `is` syntax warnings
In several places in the code e.g.,
https://github.com/climate-processes/tobac/blob/eeb574c47fceab370008b3767804ed037a269f79/tobac/tracking.py#L68
we use `it` to compare a variable to a string literal. This isn't generally good practice and we actually get `SyntaxWarning`s about it: 
`tobac/analysis.py:89: SyntaxWarning: "is not" with a literal. Did you mean "!="?`

This is an easy fix, and mostly just a reminder to myself to go through and PR a fix for this.

freemansw1:
Resolves #97 and also adds a `.gitignore` that is useful for developers (like me).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,5 @@
+*.pyc
+__pycache__
+.vscode
+htmlcov
+.coverage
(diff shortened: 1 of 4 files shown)


w-k-jones (approved):

w-k-jones:
For some reason I'd just assumed we must already have a .gitignore file!

PR #100, opened 2022-03-28
Remove setting backend to `agg` automatically

### LINKED ISSUE #95: `plotting.py` forces matplotlib backend to `agg`
In `tobac/plotting.py`, we force the Matplotlib backend to `agg`:
https://github.com/climate-processes/tobac/blob/e2af8e0c6f7e78f936c42539660bcb2b20e290c6/tobac/plotting.py#L2

This is not really good practice, especially as it's imported by default when users run `import tobac`. It's funny, @w-k-jones @galexsky and I were just talking about this. I'll get on a PR to fix this shortly.

freemansw1:
Resolves #95.

@w-k-jones I think there's a chance that this could potentially break some code. For example, if a user tries to run `tobac.plotting.plot_tracks_mask_field_loop` in a script without setting their own backend, it could break. We could resolve it by putting the matplotlib imports into each function and setting `mpl.use('agg')` there, but I still think we're better off having the users set their own backend as `agg` isn't appropriate for all cases.

### DIFF
--- a/tobac/plotting.py
+++ b/tobac/plotting.py
@@ -1,6 +1,5 @@
 import matplotlib as mpl
-
-import matplotlib.pyplot as plt
+import warnings
 import logging
 from .analysis import lifetime_histogram
 from .analysis import histogram_cellwise, histogram_featurewise
@@ -24,10 +23,15 @@ def plot_tracks_mask_field_loop(
     margin_top=0.05,
     **kwargs
 ):
+    mpl_backend = mpl.get_backend()
+    if mpl_backend != 'agg':
+        warnings.warn("When using tobac plotting functions that render a figure, you may need "
+                    "to set the Matplotlib backend to 'agg' by `matplotlib.use('agg').")
+    import matplotlib.pyplot as plt    
     import cartopy.crs as ccrs
     import os
     from iris import Constraint
-
+    
     os.makedirs(plot_dir, exist_ok=True)
     time = mask.coord("time")
     if name is None:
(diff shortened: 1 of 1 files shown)


w-k-jones (approved):

w-k-jones:
I agree, it's better to remove it to prevent unexpected knock-on effects for users. I think for now we can simply warn users that they may need to set a backend themselves (perhaps throw a warning when tobac.plotting is imported?) and deal with any problems if they appear. I think it would be best to make the plotting tools an optional import in future and remove matplotlib as a dependency (unless using plotting)

w-k-jones (approved):

w-k-jones:
> Okay, I added warnings if `agg` isn't set as the backend where figures are rendered. I've also moved the `import matplotlib.pyplot as plt` import to each individual function that uses it to give us the opportunity to set the backend there if we feel that it is more appropriate.
> 
> As I've said in a few different venues now, I do wonder what the future of the plotting module (and to a lesser extent the analysis module) is. I think there are advantages to having easy tools available for new users, but at the moment they both rely extensively on `iris` tools and both need substantial refactoring to improve ease of use and documentation. I'm going to ponder this some and raise an issue about it. If we want to start depreciating functions, v1.4/1.5 is a good opportunity to do so.

I agree, I think that at the very least we should avoid needing matplotlib as a mandatory dependency. Both the plotting and analysis modules are heavily reliant on iris, so will obviously need big changes when we shift to xarray. I think we should keep both as they are for now, but at some point add warnings to note that these tools may not be available in version 2. I think that some plotting and analysis tools should remain, but more as examples for usage of tobac and for demonstrating how e.g. figures in papers and the examples notebooks are produced than as a core part of tobac.

Anyway, happy for this PR to be merged

Issue #93, opened 2022-03-21
Speedups for calculate area


w-k-jones:
The function [calculate area](https://github.com/climate-processes/tobac/blob/e2af8e0c6f7e78f936c42539660bcb2b20e290c6/tobac/analysis.py#L266) in tobac.analysis is quite slow for large segmentation masks due to looping over each feature individually. This would be a good candidate for improving performance by vectorizing this loop using scipy.labelled_comprehension or scipy.binned_statistic

w-k-jones:
It turns out that the majority of the time it calculate_area is spend adding a column to the dataframe with the results. I guess this is good motivation to move to using xarray instead...

freemansw1:
Interesting. That's similar to what we saw in the tracking module. You may be able to get good performance by constructing the column first as a dict/list/'numpy array and then adding the column in as one vectorized operation. That has generally worked well for us.

w-k-jones:
Turns out I was wrong, and in fact I'd just not properly reloaded my changes due to relative imports and it just appeared that way, when in reality it was still running the old, slow code. Turns out it doesn't actually take 6 seconds to add a new array from a numpy array of values...

PR #102, opened 2022-03-28
Speedup to analysis.calculate_area

w-k-jones:
PR to fulfill #93. Substantial performance improvements from using scipy.ndimage.labeled_comprehension

Previous run-time:
`6.2 s ± 174 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)`

New run-time:
`11.2 ms ± 194 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)`

I haven't added lat/lon area calculation, but this could also be done fairly easily

### DIFF
--- a/tobac/analysis.py
+++ b/tobac/analysis.py
@@ -423,6 +423,7 @@ def calculate_area(features, mask, method_area=None):
     from tobac.utils import mask_features_surface, mask_features
     from iris import Constraint
     from iris.analysis.cartography import area_weights
+    from scipy.ndimage import labeled_comprehension
 
     features["area"] = np.nan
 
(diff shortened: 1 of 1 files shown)


w-k-jones:
I've added a new method to calculate area from 2D lat/lon coords. It uses a few approximations, so it's fairly inaccurate, but it'll do for now. I don't think it's worthwhile making any more updates to the analysis functions as they rely heavily on iris, and at some point we'll want to shift that over to xarray

freemansw1 (commented):
I'm happy with these changes, but I do want to try running them on our data before I approve. I should have time to do that later this week.

w-k-jones:
Great, is it ok to go ahead and merge?

freemansw1:
Yes, sorry for not being explicit about that! Happy to have this merged.

Issue #48, opened 2021-03-03
UnboundLocalError: local variable 'dxy' referenced before assignment


bernatp3rs:
Hi, 

I am trying to use TOBAC but I always get the same error:
**dxy,dt=tobac.get_spacings(Precip)
  File "/usr/local/anaconda3/envs/my_env36/lib/python3.6/site-packages/tobac/utils.py", line 450, in get_spacings
    return dxy,dt
UnboundLocalError: local variable 'dxy' referenced before assignment**

I cannot even run the following example:
https://github.com/climate-processes/tobac/blob/master/examples/Example_OLR_Tracking_model/Example_OLR_Tracking_model.ipynb

Does anybody know why is this happening?

freemansw1:
Hm, this is a bit of an odd case. I'm not sure why you would be getting this error as I don't see a path for `dxy` to be undefined. Either dxy is set here:
https://github.com/climate-processes/tobac/blob/d97517c3fd5af99defbb9fa5a9b924db37f6500d/tobac/utils.py#L436
or it is set here:
https://github.com/climate-processes/tobac/blob/d97517c3fd5af99defbb9fa5a9b924db37f6500d/tobac/utils.py#L438
and if neither `if` statement is triggered, a `ValueError` should be raised. What version of `tobac` are you using?

PR #103, opened 2022-04-05
bug fix: added missing 'raise' in get_spacing to actually raise ValueError

JuliaKukulies:
fixed the bug that was mentioned #48


freemansw1:
Great catch! We should probably address this in `dev` as well.

PR #105, opened 2022-04-06
added missing 'raise' in get_spacings to actually raise ValueError

JuliaKukulies:
Same fix as #103 for ```dev```, issue #48 should be solved with this


freemansw1 (approved):

freemansw1:
Thanks for porting this over Julia!

Issue #67, opened 2022-02-02
`min_num` and `n_min_threshold` redundant


freemansw1:
In the `main`, `dev`, and `v2.0-dev` branches of tobac feature detection, `min_num` and `n_min_threshold` are both included as parameters, while only `n_min_threshold` is used. 

My suggestion, similar to #63, is to throw a `DeprecationWarning` when users pass in a value for `min_num` and eventually throw a `ValueError` as we don't use the parameter.

freemansw1:
I should have noted in my original post that this was first raised by @galexsky.

Issue #63, opened 2022-01-18
Harmonize distance parameters in tracking functions 


freemansw1:
The presence of `v_max`, `d_min`, and `d_max` are confusing in the tracking functions at this point. We need to improve both the documentation here and how tobac rectifies these three functions. Some of this has been resolved in `v2.0-dev`, but even there it is still a little confusing. 

I've looked over the code and this is my understanding of the three parameters:

`v_max`: the maximum velocity (in units of `dxy`/`dt`) that an object can move between frames/timesteps
`d_max`: the maximum distance (in the same units as `dxy`) that an object can move between frames/timesteps
`d_min`: the minimum distance; overwrites `v_max` or `d_max`. 

So, here's my proposal:
- Ultimately get rid of two of these (I think we should keep `v_max`) as they all do the same thing and overwrite each other. 
- Throw a deprecation warning now for the two we will ultimately get rid of but keep behavior the same for now. In the documentation, strongly encourage users to use whichever one is picked and state something to the effect of "do not use xx". 
- Throw a warning now if users are using any more than one of these at a time. 
- Eventually (v2 or later perhaps?), throw an error when using the two that are depreciated

Unfortunately, I'm hesitant to break compatibility with existing code now, but I think we have a path forward.

w-k-jones:
Personally I'm happy to keep both v_ and d_max, as although they overwrite each other one may be more useful than the other for the user. However, this does highlight several related issues:

1. If both v_max and d_max are set, which should be used? I would argue that which gives the smallest limit on motion should be used, rather than simply always overwriting 1 with the other
2. If we keep both v_max and d_max, then it would make sense to also have a v_min
3. d_min, v_min shouldn't overwrite d_max/v_max, but instead set a minimum bound on the motion. i.e. if both d_max and d_min are set, then objects can only be selected that move greater than d_min and less than d_max. I'm not sure how this works with trackpy however.

freemansw1:
This has been resolved by #107.

PR #107, opened 2022-04-07
Parameter warnings

w-k-jones:
Issues #67 and #63 raised times in which duplicate or redundant parameters are used in feature detection and tracking. This PR adds future warnings for two parameters that are to be depreciated in future (min_num in feature detection and d_min in tracking) and adds an exception if multiple conflicting parameters are used. Example notebooks and tests have been updated to stop using depreciated parameters.


w-k-jones:
Thought it would be good to add depreciation warnings for the v1.3 release so that they are visible for a time before we fully remove the depreciated parameters

freemansw1 (approved):
I'm very happy with this approach, thanks for tackling it.

PR #113, opened 2022-04-08
Update example notebooks to check if data is present

w-k-jones:
Checks for existence of example data files and downloads from zenodo if they are not present


examples/Example_OLR_Tracking_satellite/Example_OLR_Tracking_satellite.ipynb
```diff
@@ -79,15 +79,20 @@
...
-    "# file_path='https://zenodo.org/record/3195910/files/climate-processes/tobac_example_data-v1.0.1.zip'\n",
-    "# tempfile='temp.zip'\n",
-    "# print('start downloading data')\n",
-    "# request=urllib.request.urlretrieve(file_path,tempfile)\n",
-    "# print('start extracting data')\n",
-    "# zf = zipfile.ZipFile(tempfile)\n",
-    "# zf.extractall(data_out)\n",
-    "# print('example data saved in')"
+    "# Download the data: This only has to be done once for all tobac examples and can take a while\n",
+    "data_path=os.path.join(data_out,'*','data','Example_input_OLR_satellite.nc')\n",
```
JuliaKukulies, line 83: Not sure how important this is, but since you are updating this part now anyhow, maybe we could also replace the `os.path` functions by using [pathlib](https://docs.python.org/3/library/pathlib.html) to go with the flow?

JuliaKukulies (approved):

examples/Example_OLR_Tracking_satellite/Example_OLR_Tracking_satellite.ipynb
w-k-jones: I'll have to look into it, as I haven't used pathlib before. Maybe it's something for a future update?

JuliaKukulies: OK, sounds good! Lets save this for another time and consider this PR approved :) 

Just for info: My understanding is that in general it could be advantageous to migrate to `pathlib,` since it is a bit easier to read and to use. For this short code snippet maybe not so relevant, but there a lot of operations that you can directly apply on `pathlib.Path` objects, whereas you would need additional libraries with `os.path`. It is also quite straight forward to convert, your code would maybe look something like: 

```suggestion

from pathlib import Path 
data_path=Path(data_out)
data_file = list(data_path.rglob('*/data/Example_input_OLR_model.nc'))
if len(data_file) == 0 :
    file_path='https://zenodo.org/record/3195910/files/climate-processes/tobac_example_data-v1.0.1.zip'
    tempfile= Path('temp.zip')
    print('start downloading data')
    request=urllib.request.urlretrieve(file_path,tempfile)
    print('start extracting data')
    shutil.unpack_archive(str(tempfile),str(data_out))
    tempfile.unlink()
    print('data extracted')
    data_file = list(data_path.glob('*'))
```

w-k-jones: Thanks for the example, I went through the example notebooks to update them to use pathlib, and it actually meant that we can replace os and glob completely as imports

examples/Example_OLR_Tracking_model/Example_OLR_Tracking_model.ipynb
```diff
@@ -79,62 +78,63 @@
...
+    "# Download the data: This only has to be done once for all tobac examples and can take a while\n",
+    "data_file = list(data_out.rglob('data/Example_input_OLR_model.nc'))\n",
+    "if len(data_file) == 0:\n",
+    "    file_path='https://zenodo.org/record/3195910/files/climate-processes/tobac_example_data-v1.0.1.zip'\n",
+    "    #file_path='http://zenodo..'\n",
+    "    tempfile=Path('temp.zip')\n",
+    "    print('start downloading data')\n",
+    "    request=urllib.request.urlretrieve(file_path, tempfile)\n",
+    "    print('start extracting data')\n",
+    "    shutil.unpack_archive(tempfile, data_out)\n",
```
JuliaKukulies, line 109: Nice, in my example I was not aware that `shutil` is now also compatible with `pathlib.Path` objects

examples/Example_OLR_Tracking_satellite/Example_OLR_Tracking_satellite.ipynb
```diff
@@ -22,13 +22,12 @@
     "import iris\n",
     "import numpy as np\n",
     "import pandas as pd\n",
-    "import os\n",
     "import matplotlib.pyplot as plt\n",
     "import iris.plot as iplt\n",
     "import iris.quickplot as qplt\n",
     "import zipfile\n",
```
JuliaKukulies, line 28:
```suggestion
    "import shutil\n",
```

examples/Example_Precip_Tracking/Example_Precip_Tracking.ipynb
```diff
@@ -30,14 +30,13 @@
     "import iris\n",
     "import numpy as np\n",
     "import pandas as pd\n",
-    "import os,sys\n",
     "import matplotlib.pyplot as plt\n",
     "import iris.plot as iplt\n",
     "import iris.quickplot as qplt\n",
     "import datetime\n",
     "import zipfile,shutil\n",
```
JuliaKukulies, line 37:
```suggestion
    "import shutil\n",
```

JuliaKukulies (changes requested):
Great, thanks for making these changes! I have noted that in one of the notebooks, we forget to import the `shutil` module and in two others, `zipfile` is still imported even though it is not needed (see below). If you could quickly fix this, I am happy to approve the PR! And it would probably be good to directly submit a PR with the same changes to `https://github.com/climate-processes/tobac-tutorials/`

freemansw1 (approved):
I'm happy with these changes in general. I agree with the suggestions made by @JuliaKukulies and think it would be best to resolve those before merging. Is there a reason that we wanted this in 1.3? Should we just go ahead and release 1.3, and once these changes (and any bugfixes) are in, we can increment to 1.3.1?

examples/Example_OLR_Tracking_model/Example_OLR_Tracking_model.ipynb
w-k-jones: Yes, we should certainly remove Python 2 support, and I'm also in favor of limiting to >=3.7. In fact the conda-forge recipe is already restricted to >=3.6, so it would be easy to bump that up. We should edit the readme to reflect this

w-k-jones:
> I'm happy with these changes in general. I agree with the suggestions made by @JuliaKukulies and think it would be best to resolve those before merging. Is there a reason that we wanted this in 1.3? Should we just go ahead and release 1.3, and once these changes (and any bugfixes) are in, we can increment to 1.3.1?

Yes, I think we can release 1.3 as is, then think about increasing the python requirement to >=3.7  for 1.3.1

JuliaKukulies:
Sounds all good to me!

examples/Example_OLR_Tracking_model/Example_OLR_Tracking_model.ipynb
JuliaKukulies: Perfect, thanks! I agree with you both on removing Python 2 support

JuliaKukulies (approved):

PR #115, opened 2022-04-12
removed redundant input parameter in linking_trackpy()

JuliaKukulies:
Just came across a small thing in `linking_trackpy()` : is it not better to remove the input parameter `field_in` because it is not used in the function or did you keep it on purpose? In this PR I have removed it from `tracking.py` and from the corresponding test functions. Please ignore, if you do not think this needs to be included before the release of version 1.3.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,26 +4,26 @@ _**Version 1.3:**_
 
 **Enhancements**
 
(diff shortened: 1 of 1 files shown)


JuliaKukulies:
That is a good point and a good alternative idea. That would be great if it is possible to check if users are passing input parameters per keyword or not; I will look into this.

PR #116, opened 2022-04-12
updated get_spacings() to calculate dxy from 1D or 2D lat-lon values, independent of their dimension order 

JuliaKukulies:
I updated `get_spacings()` in `utils/general.py` because previously this function assumed that for datasets with lon-lat coordinates, these were two-dimensional with 'lon' being the first dimension and 'lat' being the second dimension. I mistakenly introduced this with #65, and I think it is better to make the function more general so that it works for both 1D and 2D lat and lons and independent of their dimension order.

### DIFF
--- a/tobac/utils/general.py
+++ b/tobac/utils/general.py
@@ -272,7 +272,7 @@ def get_spacings(field_in, grid_spacing=None, time_spacing=None):
                 # for 1D lats and lons  
                 lat_axis, lon_axis = -1, -1 
                 
-        # re-write to corresponding axis in 2D coords (needed if dimensions like time and )
+        # re-write to corresponding axis in 2D coords (needed if other dimensions precede lats and lons)
         if lon_axis > lat_axis:
             lat_axis, lon_axis  = -2,  -1 
         elif lon_axis < lat_axis:
--- a/tobac/utils/general.py
+++ b/tobac/utils/general.py
@@ -259,8 +259,6 @@ def get_spacings(field_in, grid_spacing=None, time_spacing=None):
         # get min and max values of lats and lons
         lat1 = np.min(field_in.coord("latitude").points)
         lat2 = np.max(field_in.coord("latitude").points)
-        lon1 = np.min(field_in.coord("longitude").points)
-        lon2 = np.max(field_in.coord("longitude").points)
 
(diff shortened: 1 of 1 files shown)


w-k-jones (approved):

tobac/utils/general.py
```diff
@@ -259,14 +259,25 @@ def get_spacings(field_in, grid_spacing=None, time_spacing=None):
...
-        dlon = (lon2 - lon1) / np.shape(field_in.data)[1]
-        dlat = (lat2 - lat1) / np.shape(field_in.data)[2]
+        lat1, lat2 = map(radians, [lat1, lat2])
+
+        # for 1D lats and lons :
+        lat_axis = -1
+        lon_axis = -1
+        # for 2D lats and lons, check order 
+        if field_in.coord('latitude').points.ndim == 2:
+            if field_in.coords()[1].name() != 'latitude':
```
freemansw1, line 270: Should we just loop through all of the coordinates and find `latitude` and `longitude` that way?

JuliaKukulies: Yes, good suggestion! I changed this part in the code.

freemansw1 (approved):

freemansw1:
What is our policy on merging? Should @JuliaKukulies merge this, or should I given that I was the last reviewer?

w-k-jones:
We should probably set a policy for this. Personally I think it's best to leave it to whoever made the pull request in case they need to make any further additions/changes

freemansw1:
Agreed with that, for PRs done by maintainers. For non-maintainers, maybe after the last review, ask (and tag them) about whether they are ready to merge?

I'll leave it to @JuliaKukulies for this one.

JuliaKukulies:
Sounds good, I agree with your suggestion @freemansw1  to leave it to the last reviewer for non-maintainers, but communicate with the person if they are ready. Good that you brought that up! Sorry @w-k-jones for having merged your last PR on the notebook update- next time I will leave it to you

w-k-jones:
> Sorry @w-k-jones for having merged your last PR on the notebook update- next time I will leave it to you

No worries! It's not something we have a specific policy on yet, and often it's definitely easier for the approving reviewer to merge

Issue #94, opened 2022-03-21
Add Code Coverage to `main` and `dev`


freemansw1:
Based on a discussion with @w-k-jones and @galexsky, we should probably add code coverage to the `main` and `dev` branches to encourage more tests to be added. I'm happy to add it in the same way that we have it on `v2.0-dev`.

freemansw1:
I finally took some time to look into this. It seems that it needs someone with org privileges to enable it for other branches, so I'll table this until the org transition is completed.

JuliaKukulies:
Is the plan actually to create a github action for the automatic creation of a `codecov` report on pull requests instead of using travis CI? Then we should probably remove the `travis.yml` in `v2.0-dev`  at some point. Or what do you mean exactly to add the code coverage the same way as we have it on `v2.0-dev` ?

freemansw1:
I think that `codecov` is generated automagically by codecov.io and doesn't rely on travis-ci (and yes, we should remove that from v2.0-dev). It does, however, require integration with the organization. Although, to be completely honest, I'm not entirely sure! Once we get the org moved over, we can investigate further as we can start the setup process then.

JuliaKukulies:
OK, thanks for the clarification @freemansw1 ! It makes sense what you say about the integration with the organization. I would think that even though the code coverage reports are automatically by generated by codecov.io, you still would need to create a github action that runs `codecov` on every pull request? Anyhow, we will see how it looks like once we have the organization. I was just confused, because it sounded like there is already a working automatic `codecov` generation in `v2.0-dev` and all I could find was the one connected to the old travis CI.

freemansw1:
This has been resolved with #124 and will go in v1.3.2.

PR #120, opened 2022-04-13
added pull_request_template.md in .github/

JuliaKukulies:
Added a pull request check list that automatically will appear in the pull request description field. Please feel free to add any checks that I have missed or remove redundant instructions. I have put the file here to not overpopulate the root directory. We could, however also move it to doc/

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,12 @@
+* [ ] Have you followed our guidelines in CONTRIBUTING.md? 
+* [ ] Have you self-reviewed your code and corrected any misspellings? 
+* [ ] Have you written documentation that is easy to understand?
+* [ ] Have you written descriptive commit messages? 
+* [ ] Have you added NumPy docstrings for newly added functions? 
+* [ ] Have you formatted your code using black? 
+* [ ] If you have introduced a new functionality, have you added adequate unit tests?
+* [ ] Have all tests passed in your local clone? 
+* [ ] If you have introduced a new functionality, have you added an example notebook?
+* [ ] Have you kept your pull request small and limited so that it is easy to review? 
+* [ ] Have any dependent changed been merged in downstream modules of this branch? 
+
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -8,5 +8,5 @@
 * [ ] Have all tests passed in your local clone? 
 * [ ] If you have introduced a new functionality, have you added an example notebook?
 * [ ] Have you kept your pull request small and limited so that it is easy to review? 
-* [ ] Have any dependent changes been merged in downstream modules of this branch? 
+* [ ] Have the newest changes from this branch been merged? 
 
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -8,5 +8,5 @@
 * [ ] Have all tests passed in your local clone? 
 * [ ] If you have introduced a new functionality, have you added an example notebook?
 * [ ] Have you kept your pull request small and limited so that it is easy to review? 
-* [ ] Have any dependent changed been merged in downstream modules of this branch? 
+* [ ] Have any dependent changes been merged in downstream modules of this branch? 
 


JuliaKukulies:
This should of course go into `dev`, sorry for the mess. I rebased my branch onto `dev`, but can still not change it the target branch here.

freemansw1:
I was able to change the target branch to `dev`.

.github/pull_request_template.md
```diff
@@ -0,0 +1,12 @@
+* [ ] Have you followed our guidelines in CONTRIBUTING.md? 
+* [ ] Have you self-reviewed your code and corrected any misspellings?
```
freemansw1, line 2: This one is funny given the many spellings of "indices" in the original code

.github/pull_request_template.md
```diff
@@ -0,0 +1,12 @@
+* [ ] Have you followed our guidelines in CONTRIBUTING.md? 
+* [ ] Have you self-reviewed your code and corrected any misspellings? 
+* [ ] Have you written documentation that is easy to understand?
+* [ ] Have you written descriptive commit messages? 
+* [ ] Have you added NumPy docstrings for newly added functions? 
+* [ ] Have you formatted your code using black? 
+* [ ] If you have introduced a new functionality, have you added adequate unit tests?
```
freemansw1, line 7: Should we ask people to generate a coverage report? I guess this goes back to #94 , though.

.github/pull_request_template.md
```diff
@@ -0,0 +1,12 @@
...
+* [ ] Have all tests passed in your local clone? 
+* [ ] If you have introduced a new functionality, have you added an example notebook?
+* [ ] Have you kept your pull request small and limited so that it is easy to review? 
+* [ ] Have any dependent changes been merged in downstream modules of this branch?
```
freemansw1, line 11: I'm not sure what this means. Does it mean that your changes don't break other code?

freemansw1 (changes requested):

.github/pull_request_template.md
JuliaKukulies: Haha, I know. Also given that I had several misspellings in this original PR. Would you remove it?

.github/pull_request_template.md
JuliaKukulies: I think that is a good idea as the testing part really is in the end the most important of all these checks

.github/pull_request_template.md
freemansw1: I think this is a great policy to have, actually. I'm 100% on board with requiring this.

.github/pull_request_template.md
freemansw1: Sorry, I should specify that my question here is specifically on line 11.

JuliaKukulies: I guess what I meant here was to avoid merge conflicts by making sure to pull the latest changes on that branch etc. Even if git automatically manages this, maybe it is good to encourage that users think about it before submitting a PR?  I dont know - we can either re-formulate this or I am also happy to just remove this point.

freemansw1: Maybe "Have the newest changes from this branch been merged?". I suppose, as you mention, that it isn't really required, but it is probably good practice to avoid unforeseen issues (and conflicts that git doesn't detect).

.github/pull_request_template.md
freemansw1: I'm going to suggest leaving user-generated coverage reports out for now. I think we should be able to do it automatically with #94.

JuliaKukulies: OK, thanks for the update! I was actually about to raise the question if it is not better to have automated code coverage reports instead of asking the users, so I agree with you -  we will leave it here.

.github/pull_request_template.md
JuliaKukulies: Sounds good, I have changed it according to your suggestion. Besides good practice it could also be helpful for the reviewer to be aware of unmerged changes, in case this bullet point is not checked.

freemansw1 (approved):
I'm happy with these now, thanks!

JuliaKukulies:
Please go ahead @freemansw1

PR #124, opened 2022-04-24
Add code coverage via Github Actions

### LINKED ISSUE #94: Add Code Coverage to `main` and `dev`

freemansw1:
Resolves #94 and adds code coverage via codecov.io and GitHub Actions. 

We will want to merge this as a squash merge.


freemansw1:
Apologies for spamming your email with the failed attempts here; this ended up being nontrivial.

freemansw1:
Note that we can turn those coverage comments off (I like them, and am happy for them to be on).

JuliaKukulies (approved):
Nice, well done @freemansw1! It looks all good to me. Yes, probably a good idea to squash the commits. I also like the comments and think it could be useful to leave them on. I assume that the issue with the missing base report will be automatically fixed when the PR is merged?

freemansw1:
Thanks @JuliaKukulies! Yes, it should be the comparison version, but since this has never been run before, there's nothing to compare to.

freemansw1:
I am inclined to go ahead and merge this after #141 is resolved and 1.3.1 is out, unless you think we should try to get it in for 1.3.1 @JuliaKukulies?

JuliaKukulies:
I think that sounds like a good plan, especially when there will be an extra bug fix release 1.3.2, it can be released together with #140 as you suggested (and I think it would make sense and we do not need to wait for #140 to be reviewed to release 1.3.1).

Issue #121, opened 2022-04-14
`map_tracks` doesn't drop untracked cells


freemansw1:
As a consequence of allowing a user-set value for untracked cells instead of NaN, `map_tracks` now plots the untracked cells by default. 

This was first raised in an e-mail to me from Sid Gupta at BNL.

freemansw1:
This has been resolved with #130

PR #130, opened 2022-05-11
Fix `map_tracks` with untracked cell number change

### LINKED ISSUE #128: The Example_OLR_Tracking_satellite notebook fails with latest `main`
Hi,
The `examples/`Example_OLR_Tracking_satellite.ipynb` file no longer produces the expected output.
This section of code:
```
# Plot map with all individual tracks:
import cartopy.crs as ccrs
fig_map,ax_map=plt.subplots(figsize=(10,10),subplot_kw={'projection': ccrs.PlateCarree()})
ax_map=tobac.map_tracks(Track,axis_extent=axis_extent,axes=ax_map)
```

Produces a map in which one of the trajectories has gone haywire:
download

I also see that the features returned by `tobac.feature_detection_multithreshold` and then subsequently by `tobac.segmentation_2D` are identical, not sure if that's intended.

I'm running a fully updated conda environment with python 3.9 and the latest tobac version on the `main` branch.

freemansw1:
This resolves #128 and #121, related to the fact that `map_tracks` in the plotting module depends on the untracked cells being labeled as `np.nan`.

Given the number of bug reports related to this, I recommend that we release this as a hotfix 1.3.1 release.

### DIFF
--- a/tobac/plotting.py
+++ b/tobac/plotting.py
@@ -1730,8 +1730,10 @@ def plot_mask_cell_track_static_timeseries(
         plt.clf()
 
 
-def map_tracks(track, axis_extent=None, figsize=None, axes=None, untracked_cell_value = -1):
-    '''Maps the tracks of all tracked cells
+def map_tracks(
+    track, axis_extent=None, figsize=None, axes=None, untracked_cell_value=-1
+):
+    """Maps the tracks of all tracked cells
 
     Parameters
     ----------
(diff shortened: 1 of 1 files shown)


tobac/plotting.py
```diff
@@ -1730,8 +1730,45 @@ def plot_mask_cell_track_static_timeseries(
...
+    ------
+    Raises a `ValueError` if `axes` is not passed in.
+
+    """
+    if figsize is not None:
+        warnings.warn(
+            "figsize is depreciated as this function does not create its own figure.",
+            DeprecationWarning,
+        )
+    if axes is None:
```
JuliaKukulies, line 1765: Is there a reason why we set `None` as a  default value for the `axes` parameter and then raise an error when no axis object is provided? I do not see the meaning right now, so maybe it is a leftover for when the function created its own figure?

JuliaKukulies (approved):
Looks good and agree with the hot fix. Just a small comment/question which I suspect is a leftover from previously. Not sure if we want to fix this now, so I am also happy to just approve this PR.

tobac/plotting.py
freemansw1: The ultimately correct way to do this would be to have `axes` as a required positional argument, but I believe this strange structure is leftover from when the function created its own figure. I wanted to add these in there as, without these errors, it will fail in a strange way. 

Given our headaches with `field_in` and the overall uncertain future of the plotting module, I am hesitant to move `axes` to being a required positional argument for now. I think it's something we should look into in the future as we figure out what we want to do with the plotting module long term.

JuliaKukulies: OK yes, sounds good! So from my side: ready to merge

freemansw1:
I'm planning on merging this tomorrow given how small the change is, without waiting on the second reviewer.

JuliaKukulies:
Fine from my side!

PR #131, opened 2022-05-12
Support interp kind

zxdawn:
Support setting interpolation kind of coord.

### DIFF
--- a/tobac/themes/tobac_v1/feature_detection.py
+++ b/tobac/themes/tobac_v1/feature_detection.py
@@ -30,6 +30,7 @@ def feature_detection_multithreshold(
     min_num=0,
     target="maximum",
     position_threshold="center",
+    coord_interp_kind='linear',
     sigma_threshold=0.5,
     n_erosion_threshold=0,
     n_min_threshold=0,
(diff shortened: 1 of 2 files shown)


tobac/themes/tobac_v1/feature_detection.py
```diff
@@ -63,6 +64,13 @@ def feature_detection_multithreshold(
         Flag choosing method used for the position of the tracked
         feature. Default is 'center'.
 
+    coord_interp_kind : str, optional
+        The kind of interpolation as a string. Default is 'linear'.
```
freemansw1, line 68: I think this should be clearer as to what the interpolation is touching- just add something like "for coordinates" in here.

freemansw1 (changes requested):
Minor change on clarity, otherwise I'm happy with this!

zxdawn:
@JuliaKukulies Yes, please feel free to merge it ;)

JuliaKukulies:
Or actually, I will leave it to @freemansw1 who still needs to approve your changes :) but I guess they will!

PR #134, opened 2022-06-02
Post processing: Merge/Split of tracked cells.  

kelcyno:
This branch introduces the merge/split framework. This also introduces the standardized dataset in which the track, segmentation mask, and feature datasets are combined. Below is an example of the merge/split call, the standardized track dataset usage, and an example of these two datasets being merged.  We include in the merge_split.py file a quick compression, which is also shown in the example blow.

d = merge_split(Track)
ds = standardize_track_dataset(Track, refl_mask, data['ProjectionCoordinateSystem'])
both_ds = xr.merge([ds, d],compat ='override')
both_ds = compress_all(both_ds)


freemansw1:
Thanks for submitting this PR, @kelcyno . It looks like you reverted the split/merge changes, so they don't appear in here. Also, it looks like the isolation changes from #76 are in here too; was that intentional, and is it needed for splits/merges to work? If so, happy to have these combined, but I'll close #76. 

I've waited to assign reviewers to this PR for now. I'll likely assign myself and @JuliaKukulies to review this, ultimately.

kelcyno:
@freemansw1 I believe I originally pushed on the wrong branch, then reverted. I just pushed the merge_split.py file, this is the only file needed for the merge/split, and the isolation changes are not needed for this to work.

freemansw1:
Ah, makes sense. Thanks for updating it. Given that the isolation changes aren't needed for this, my suggestion is to remove them from this PR and we can continue our discussion in #76, if that's okay with you. In the meantime, I've requested @JuliaKukulies and myself to review the PR, but I will wait to review until the isolation changes are removed (and I have more time). 

Also- I'm not entirely sure why the Github actions aren't running here. I am a bit short on time at the moment, but I'll try to investigate in the next week or two.

kelcyno:
Alright - I updated the segmentation file to remove the isolation changes, using the default segmentation.py file. If this doesn't go smoothly let me know and I'll try deleting the branch and redoing the pull request with just the merge_split.py file.

freemansw1:
Hm- I still see the segmentation changes in there. Can you try pushing again with them removed? Also, can you remove the file in the Build directory?

Issue #133, opened 2022-05-25
subnet_size in linking_trackpy() does not work with adaptive search


snilsn:
Hi all,

this may be very specific, but combining _subnet_size_ with adaptive_stop and _adaptive_step_ in the _linking_trackpy()_-function does not work. The reason is that _subnet_size_ specifies 

**trackpy.linking.Linker.MAX_SUB_NET_SIZE**, 

but when adaptive search is used, 

**trackpy.linking.Linker.MAX_SUB_NET_SIZE_ADAPTIVE**

 is the upper limit of the subnetwork size. This value is 15 by default, while tobac sets trackpy.linking.Linker.MAX_SUB_NET_SIZE to 1 if not specified otherwise.

For reference: https://soft-matter.github.io/trackpy/v0.3.0/tutorial/adaptive-search.html (the word of caution in the fourth paragraph)

freemansw1:
Thanks for raising this issue. It does, at first glance, appear as a bug to me. I think the core developers are stretched a bit thin at the moment, so addressing this from our end will probably take some time. If you (or someone else) is interested, we'd happily entertain a PR on this, though!

snilsn:
Okay, I will try to fix this

PR #140, opened 2022-06-20
Correct parameter for the subnet size in adaptive search

snilsn:
This pull request aims at solving issue #133.

If neither adaptive_step nor adaptive_stop are given, **MAX_SUB_NET_SIZE** is set, otherwise **MAX_SUB_NET_SIZE_ADAPTIVE** is changed in the _linking_trackpy()_-function.


freemansw1:
I've requested @JuliaKukulies and I as reviewers for this. I should be able to get to this next week. 

Once this is reviewed, it may make sense to release it as part of a 1.3.2 bugfix release. I'm hesitant to try to get it into 1.3.1, but I could see this and #124 for example making it into a 1.3.2 release. For now, I've kept it on the 1.5 milestone, but may move it to a new 1.3.2 milestone.

JuliaKukulies (commented):
Thanks @snilsn, very good and careful observation! No objection from my side for the code changes you made. The only thing I wonder is whether we can add a more clear description for the `subnetwork_size` parameter in the docstrings? As I understand it, you would want it to be rather small when you do adaptive search and rather large when you do not. So this is important information the user needs to know.

freemansw1 (approved):
I'm happy with these changes. Good catch on finding the bug.

freemansw1:
With the approving reviews, @snilsn please confirm that you're ready to merge and we will get this in `dev` with the intent to release as part of v1.3.2.

snilsn:
Yes, from my side everything is ready for merging.

> The only thing I wonder is whether we can add a more clear description for the `subnetwork_size` parameter in the docstrings

I will adress this in an update of PR #138

JuliaKukulies (approved):

JuliaKukulies:
Good, yes agreed that the documentation update is better for #138 ! Thanks!

PR #147, opened 2022-06-27
Added depreciation warnings to two segmentation functions

### LINKED ISSUE #145: Possibly obsolete functions/wrappers in segmentation.py
Hello everybody,

@JuliaKukulies and me have independently stumbled upon the functions `segmentation_3D()`, `segmentation_2D()`, `watershedding_3D()` and `watershedding_2D()` in _segmentation.py_. 

These don't seem to have any real use, since they only call the `segmentation()` function itself, and, in case of the watershedding ones, set the method to watershedding, which is the default anyway. Writing docstrings for them feels a bit redundant. Maybe these can be removed in a future release?

In v2 they are already gone, so this would also be a step towards compatibility.

On the other hand, I just noticed that the functions are used in all the example notebooks of v1, so at least those would need to be adjusted. 

What are your thoughts on that?

freemansw1:
This resolves #145 and calls for the `segmentation_2D` and `segmentation_3D` to be removed by v1.5.0. 

### DIFF
--- a/tobac/segmentation.py
+++ b/tobac/segmentation.py
@@ -310,10 +310,33 @@ def segmentation(
 
 
 def watershedding_3D(track, field_in, **kwargs):
+    warnings.simplefilter(
+        "always", DeprecationWarning
+    )  # turn off filter; this is an easy switch for users.
+    warnings.warn(
+        "Warning: using deprecated function `watershedding_3D`. Switch to `segmentation` instead."
+        " The arguments and output of `segmentation` is the same, so it is an easy switch."
+        " This function will be removed in v1.5.0.",
+        category=DeprecationWarning,
+        stacklevel=2,
+    )
+    warnings.simplefilter("default", DeprecationWarning)  # reset filter
+
     kwargs.pop("method", None)
     return segmentation_3D(track, field_in, method="watershed", **kwargs)
 
 
 def watershedding_2D(track, field_in, **kwargs):
+    warnings.simplefilter(
+        "always", DeprecationWarning
+    )  # turn off filter; this is an easy switch for users.
+    warnings.warn(
+        "Warning: using deprecated function `watershedding_2D`. Switch to `segmentation` instead."
+        " The arguments and output of `segmentation` is the same, so it is an easy switch."
+        " This function will be removed in v1.5.0.",
+        category=DeprecationWarning,
+        stacklevel=2,
+    )
+
     kwargs.pop("method", None)
     return segmentation_2D(track, field_in, method="watershed", **kwargs)
--- a/tobac/segmentation.py
+++ b/tobac/segmentation.py
@@ -1,4 +1,5 @@
 import logging
+from multiprocessing.sharedctypes import Value
 import warnings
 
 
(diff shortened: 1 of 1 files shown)


JuliaKukulies (approved):
Happy with this!

freemansw1:
As noted in my review of PR #138 and in the original issue #145, I've missed the two `watershed` functions. I'll update this PR and re-request a review. Also, I'm not sure why the codecov.io bot isn't working. I'll open another issue for that, although the problem may simply be that the code coverage stuff isn't on main yet.

tobac/segmentation.py
```diff
@@ -286,10 +310,33 @@ def segmentation(
...
     return segmentation_3D(track, field_in, method="watershed", **kwargs)
 
 
 def watershedding_2D(track, field_in, **kwargs):
+    warnings.simplefilter(
+        "always", DeprecationWarning
+    )  # turn off filter; this is an easy switch for users.
+    warnings.warn(
+        "Warning: using deprecated function `watershedding_2D`. Switch to `segmentation` instead."
+        " The arguments and output of `segmentation` is the same, so it is an easy switch."
```
JuliaKukulies, line 335: With the tiny difference that for `watershedding_2D` and `watershedding_3D` `dxy` is a key word argument. Do we need to mention that or not?

JuliaKukulies (approved):
True! Note that both warnings appear now when one of the watershedding function is called, because `watershedding_2D` returns `segmentation_2D`, but I think it makes most sense this way. Just a small comment on the warning text, which I am not sure how important it is, so you may as well ignore this comment.

tobac/segmentation.py
freemansw1: I think I have resolved this with the new commits- `watershedding_2D` and `watershedding_3D` now call `segmentation` directly, and if `dxy` isn't included, a `ValueError` is raised, as it is needed down the line.

tobac/segmentation.py
```diff
@@ -323,7 +324,9 @@ def watershedding_3D(track, field_in, **kwargs):
     warnings.simplefilter("default", DeprecationWarning)  # reset filter
 
     kwargs.pop("method", None)
-    return segmentation_3D(track, field_in, method="watershed", **kwargs)
+    if 'dxy' not in kwargs:
+        raise ValueError("Need grid spacing `dxy`.")
+    return segmentation(track, field_in, dxy=kwargs.get('dxy'), method="watershed", **kwargs)
```
JuliaKukulies, line 332: This will throw a `TypeError` because `segmentation()` gets multiple values for keyword argument `dxy` (one time defined via `dxy=kwargs.get('dxy')` and then again through `**kwargs`. Maybe this could be an alternative? 

```
if 'dxy' not in kwargs:
        raise ValueError("Need grid spacing `dxy`.")

dxy = kwargs.get("dxy")
kwargs.pop("dxy", None)    

return segmentation(track, field_in, dxy=dxy, method="watershed", **kwargs)
```
Or you have a better idea :)

JuliaKukulies (changes requested):
Thanks @freemansw1! I think that was smart the way you solved this. We just need to make a small change to the code, because now `dxy` is double defined in the input to the segmentation function. This raises an error even though it is of course the same value for `dxy`. Have a look at my suggestion

tobac/segmentation.py
freemansw1: Great suggestion, and great catch!

JuliaKukulies (approved):
Alright, happy to merge this now!

PR #154, opened 2022-07-04
Revert "Added depreciation warnings to two segmentation functions"

freemansw1:
Reverts tobac-project/tobac#147 until we resolve the best way to do this as discussed in #145.

### DIFF
--- a/tobac/segmentation.py
+++ b/tobac/segmentation.py
@@ -1,6 +1,4 @@
 import logging
-from multiprocessing.sharedctypes import Value
-import warnings
 
 
 def segmentation_3D(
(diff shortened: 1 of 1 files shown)


JuliaKukulies (approved):
Reversal of depreciation warnings approved! Does the codecov check fail because the coverage decreases when you remove a bunch of lines or is there are different reason for this? If so, what is our threshold for this?

freemansw1:
I do believe that `codecov` fails because I reduce code coverage. I think at this stage in our development, I'm generally okay with reducing code coverage some (this is -0.06%), but yeah this is something we should talk about more broadly. I think we should probably work on updating our contributing guidelines at some point to formalize some of these processes.

PR #157, opened 2022-07-09
Merge `dev` into `main` for v1.3.2

freemansw1:
Bringing in our `dev` changes into `main` for v1.3.2. 

Question for the group: do we want to keep doing these as PRs or should we just merge `dev` (or our new hotfix branch) into `main`?

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,3 @@
+[run]
+source=tobac
+omit = tobac/tests/*
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,12 @@
+* [ ] Have you followed our guidelines in CONTRIBUTING.md? 
+* [ ] Have you self-reviewed your code and corrected any misspellings? 
+* [ ] Have you written documentation that is easy to understand?
+* [ ] Have you written descriptive commit messages? 
+* [ ] Have you added NumPy docstrings for newly added functions? 
+* [ ] Have you formatted your code using black? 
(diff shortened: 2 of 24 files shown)


JuliaKukulies (approved):

JuliaKukulies:
> Question for the group: do we want to keep doing these as PRs or should we just merge `dev` (or our new hotfix branch) into `main`?

Maybe keep doing PRs at least until we have figured out the best way of managing release branches? It is probably good to have this extra check before merging into `main` when we have multiple release branches from `dev` (e.g. to assure that we are merging the right version and that the release branch has all the features it should have). 

I am thinking that instead we could skip the PR for version bump and changelog update by automatizing 
- the commit and push of these two steps to release branch
- create a PR to merge into `main` 
- create a PR to merge back latest release into `dev` (and/or other release branches)

or do you think that this would not necessarily make things less complicated because we still need something to trigger these actions?

freemansw1:
Thanks for the thoughts @JuliaKukulies! Agreed, but I'm not sure that those need to be set up as actions (unless that's reasonably easy to do?). Maybe we just do a commit to the appropriate repo bumping the version and with the changelog, that way we only have to review one PR for merging into `main`?

Are you happy for me to merge and release?

JuliaKukulies:
Yes, please go ahead and merge @freemansw1! 

I am actually also not so sure if it gives us anything to have these as actions, but I do think it is rather easy to set up. But I think your suggestion sounds like a reasonable plan! Better than the other way around (PR for changelog and version bump and then just commit the merge), because we keep that last check of what we merge.

PR #162, opened 2022-07-19
added google groups links

RBhupi:


freemansw1:
Note that I changed the target here to be the 1.4.0 release candidate, which results in the changes to `main` to be merged into `RC_v1.4.0` branch.

RBhupi:
Yeah. I thought of adding these links after the SciPy talk.
I think the lists are evolving differently than we initially thought.
We also have https://groups.google.com/g/tobac-developers group.

RBhupi:
I can open the tobac group for all to join and delete or hide tobac-users group if that sounds good. Let's talk about it at the next meeting.

JuliaKukulies:
I was not even aware that we had these two different groups to be honest, but I think your suggestion to get rid of tobac-users sounds good @freemansw1 @RBhupi. 

> and then create a mailing list for developers specifically.

I think we already have a google group/mailing list for developers only: https://groups.google.com/g/tobac-developers/about ? Would it not be easier if we stick to Slack instead of communicating on different platforms? But I am not sure how well the email integration in Slack works, as it is certainly useful to be able to send an email to a specific subgroup or let users contact us via mail if they do not have a Slack account.

CHANGELOG.md
```diff
@@ -1,5 +1,18 @@
...
+
+- Fixed a bug with Feature Detection that caused it to fail when using `weighted_abs` position [#148](https://github.com/tobac-project/tobac/pull/148)
+- Fixed a bug where adaptive_search within `linking_trackpy` was not working correctly [#140](https://github.com/tobac-project/tobac/pull/140)
+
+**Repository enhancements**
+
+- Added automatic code coverage reports [#124](https://github.com/tobac-project/tobac/pull/124)
+- Added automatic building of readthedocs documentation on pull requests
+
+
```
JuliaKukulies, line 15: Note that these changes from release 1.3.2 have not been merged back yet

JuliaKukulies (commented):
@freemansw1 It is probably best if we merge back v1.3.2 into RC_v.1.4.0 and RC_v.1.5.0 as soon as possible, or what do you think?

CHANGELOG.md
freemansw1: Yes, agreed. We should merge 1.3.2 back into the RC branches. I think we should be able to do this without a PR, are you happy if I just make a merge commit?

freemansw1:
> Would it not be easier if we stick to Slack instead of communicating on different platforms? But I am not sure how well the email integration in Slack works, as it is certainly useful to be able to send an email to a specific subgroup or let users contact us via mail if they do not have a Slack account.

I like Slack, but I think people really have to buy in to use it. I agree that we should cross-post everything there (which I haven't been great about!), but I agree we need to keep e-mail as it's a much lower barrier of entry.

CHANGELOG.md
JuliaKukulies: Yes please go ahead :) Thanks!

JuliaKukulies:
Ok yes, I agree to keep the mailing lists. But good point from @RBhupi that we already have the one for developers, so probably no need to create another.

freemansw1:
After discussion in the developers' meeting today, I think the plan is to stick with the developers and main `tobac` mailing lists and get rid of the users one. I think we should also advertise the slack in the readme.

freemansw1:
@RBhupi I think after discussion today, the plan is to advertise the two mailing lists, but not to put the slack link on the readme given spam concerns. We'll instead start distributing the slack link with releases, beginning with v1.4 (because I forgot today). 

So, that just leaves switching around the mailing lists. Are you happy to update the PR, or do you want us to do so?

freemansw1:
OK, I updated the mailing list links and added a brief slack discussion. Re-requesting review from @JuliaKukulies and requesting review from @snilsn.

JuliaKukulies (approved):
Excellent, thanks @freemansw1 !

RBhupi:
Thanks Sean!
The other group with only a few members is now deactivated.

snilsn (approved):
All good @freemansw1, approved from my side

Issue #169, opened 2022-08-09
method = 'predict' is not working


snilsn:
Another bug related to the linking process:
`method=predict` of `linking_trackpy()` is not working at the moment.

It's a problem of `trackpy`, I have documented my findings in a comment here: 

https://github.com/soft-matter/trackpy/issues/699

I think we have three options:

1. Wait for a fix in trackpy. That might take a while though, the issue hasn`t been touched in 2 months
2. Fix the problem in trackpy ourselves
3. Implement the workaround I found with `link_df_iter` in tobac

I would probably go with the third option.

freemansw1:
This explains some of the strange behavior I have noticed recently. Yes, we should fix this ASAP. I'm happy for us to implement a workaround; what is the workaround that you have proposed?

freemansw1:
Long term, as you note, we should probably work on a fix upstream. They were pretty receptive to my last PR, but it did take a while to work through.

snilsn:
> what is the workaround that you have proposed?

The implementation in `tobac` looks like this:

```
pred = tp.predict.NearestVelocityPredict(span=1)
trajectories_unfiltered = pred.link_df(
            features_linking,
            search_range=search_range,
            memory=memory,
            pos_columns=["hdim_1", "hdim_2"],
            t_column="frame",
            neighbor_strategy="KDTree",
            link_strategy="auto",
            adaptive_step=adaptive_step,
            adaptive_stop=adaptive_stop
        )
```
We could avoid the usage of `link_df` by generating an iterable of the frames and use 'link_df_iter' instead:
```
features_linking_list = [frame for i, frame in features.groupby('frame')]
pred = tp.predict.NearestVelocityPredict(span=1)
trajectories_unfiltered = pred.link_df_iter(
            features_linking_list,
            search_range=search_range,
            memory=memory,
            pos_columns=["hdim_1", "hdim_2"],
            t_column="frame",
            neighbor_strategy="KDTree",
            link_strategy="auto",
            adaptive_step=adaptive_step,
            adaptive_stop=adaptive_stop
trajectories_unfiltered = pd.concat(trajectories_unfiltered)
```
There probably are smarter and more efficient ways to do this, that's just how I got it to actually predict something quick and dirty.

snilsn:
> Long term, as you note, we should probably work on a fix upstream.

Yes, and the issue I commented on in trackpy already proposes something, but that didn't fix it for me.

[...]

PR #170, opened 2022-08-24
Implementing Workarounds and Test for method='prediction'

snilsn:
This PR aims at solving #169 temporarily by implementing two workarounds for problems upstream in trackpys predictive tracking in `linking_trackpy()`:

1. Switching to `link_df_iter` from `link_df` since the latter method isn't working correctly.
2. Avoiding to set `pos_columns` in `link_df_iter`, because this causes incorrect tracking.

The issue in trackpy for reference: https://github.com/soft-matter/trackpy/issues/699

Furthermore a test to check whether prediction works correctly is added

not needed: 

### DIFF
--- a/tobac/tracking.py
+++ b/tobac/tracking.py
@@ -117,7 +117,9 @@ def linking_trackpy(
         features.rename(columns={"hdim_1": "y", "hdim_2": "x"}, inplace=True)
 
         # generate list of features as input for df_link_iter to avoid bug in df_link
-        features_linking_list = [frame for i, frame in features.groupby("frame", sort=True)]
+        features_linking_list = [
+            frame for i, frame in features.groupby("frame", sort=True)
+        ]
 
         pred = tp.predict.NearestVelocityPredict(span=1)
         trajectories_unfiltered = pred.link_df_iter(
--- a/tobac/tracking.py
+++ b/tobac/tracking.py
@@ -113,6 +113,9 @@ def linking_trackpy(
         )
     elif method_linking == "predict":
 
+        # avoid setting pos_columns by renaimng to default values to avoid trackpy bug
+        features.rename(columns={"hdim_1": "x", "hdim_2": "y"}, inplace=True)
+
         # generate list of features as input for df_link_iter to avoid bug in df_link
         features_linking_list = [frame for i, frame in features.groupby("frame")]
 
@@ -121,7 +124,7 @@ def linking_trackpy(
             features_linking_list,
             search_range=search_range,
             memory=memory,
-            pos_columns=["hdim_1", "hdim_2"],
+            #pos_columns=["hdim_1", "hdim_2"], # not working atm
             t_column="frame",
             neighbor_strategy="KDTree",
             link_strategy="auto",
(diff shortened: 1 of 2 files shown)


freemansw1:
Thanks for putting this in so quickly, @snilsn . Given her work on this issue, I've asked @kelcyno to review along with @JuliaKukulies and myself. Hopefully we can get this out next week as a hotfix release. 

I was talking to someone about this issue today, and they raised a good point- users may want to reduce the `search_range` parameter. It's possible that it should be lower than it is in many users' scripts now.

tobac/tracking.py
```diff
@@ -113,12 +113,18 @@ def linking_trackpy(
         )
     elif method_linking == "predict":
 
+        # avoid setting pos_columns by renaimng to default values to avoid trackpy bug
+        features.rename(columns={"hdim_1": "x", "hdim_2": "y"}, inplace=True)
+
+        # generate list of features as input for df_link_iter to avoid bug in df_link
+        features_linking_list = [frame for i, frame in features.groupby("frame")]
```
freemansw1, line 120: Is this guaranteed to be sorted? If not, should `features_linking_list` be sorted by `frame`?

tobac/tracking.py
```diff
@@ -113,12 +113,18 @@ def linking_trackpy(
         )
     elif method_linking == "predict":
 
+        # avoid setting pos_columns by renaimng to default values to avoid trackpy bug
+        features.rename(columns={"hdim_1": "x", "hdim_2": "y"}, inplace=True)
```
freemansw1, line 117: I'm not sure that it matters as long as it's consistent, but I believe that elsewhere in the code, `hdim_1` is typically the `y` dimension. But I suppose that it would be the other if users input an array that had dimensions `nx, ny`? I'm happy with this as-is.

freemansw1 (commented):
Generally happy with this but want to make sure that we don't end up with sorting issues. 

As a note, I checked that the test fails without the workaround, and it does, which is good.

tobac/tracking.py
kelcyno: I recommend swapping as Sean mentioned for the sake of consistency.

kelcyno (commented):
Both the test and tracking updates are sufficient. Thanks for submitting @snilsn

tobac/tracking.py
snilsn: Yes, `sort=True` is the default for `groupby`. We don`t have to worry about this, but I think I`ll make it explicit still, just in case anyone else is wondering.

tobac/tracking.py
snilsn: Ok, let's keep this consistent then.

snilsn:
Thanks @kelcyno and @freemansw1 for looking through this so quickly.

Have you noticed a change in performance for larger datasets with these changes?

I'm thinking about adding another assert statement to the test that checks if the output with `method='predict'` is different from `method='random'`. That would ensure that prediction actually has an influence on the linking.

tobac/tracking.py
freemansw1: Thanks. Pandas has shown that they are completely okay with changing defaults with little warning, so this makes me feel better.

JuliaKukulies (approved):
This workaround looks OK to me, many thanks for this PR @snilsn! I agree with releasing this as a hot fix asap. That was a good catch -  I think we used the "predict" parameter in nearly all our examples. I tested it with an input dataframe with ~ 300 000 features and the time needed for the linking increased by about 60 %. I think adding an assert statement to the test that assures that the output of the two linking method is a good idea.

snilsn:
I added the second assert to the test.

@JuliaKukulies Thank for checking the performance. This probably the effect of the prediction actually happening, but reducing the `search_range` like @freemansw1 suggested could maybe help with that.

freemansw1 (approved):

kelcyno (approved):

PR #175, opened 2022-09-09
Adding version numbers 

### LINKED ISSUE #166: `tobac.__version__` should report version
Like many other packages, we should report our version number when calling `__version__` from the main `tobac` package. It looks like there are canonical ways to do this here: https://packaging.python.org/en/latest/guides/single-sourcing-package-version/

freemansw1:
This resolves #166 and has tobac report its version number consistent with `setup.py`. 

### DIFF
--- a/tobac/tests/test_import.py
+++ b/tobac/tests/test_import.py
@@ -8,14 +8,14 @@ def test_dummy_function():
 
 def test_version():
     """Test to make sure that we have a version number included.
-    Also test to make sure that the version number complies with 
+    Also test to make sure that the version number complies with
     semantic versioning guidelines.
     If it's not, this should result in an error.
     """
     import re
+
     assert type(tobac.__version__) is str
     # Make sure that we are following semantic versioning
-    # i.e., our version is of form x.x.x, where x are all 
+    # i.e., our version is of form x.x.x, where x are all
     # integer numbers.
     assert re.match(r"[0-9]+\.[0-9]+\.[0-9]+", tobac.__version__) is not None
-


setup.py
```diff
@@ -1,8 +1,33 @@
...
+
+
+def get_version(rel_path):
+    for line in read(rel_path).splitlines():
+        if line.startswith("__version__"):
+            delim = '"' if '"' in line else "'"
+            return line.split(delim)[1]
+    else:
+        raise RuntimeError("Unable to find version string.")
+
```
JuliaKukulies, line 26:
```suggestion
PACKAGE_NAME= "tobac"
```

setup.py
```diff
@@ -1,8 +1,33 @@
...
+
 setup(
     name="tobac",
```
JuliaKukulies, line 29:
```suggestion
    name=PACKAGE_NAME,
```

JuliaKukulies: a suggestion to define a constant for the package name because "tobac" occurs now 3 times in `setup.py` (see l.30, 50)

setup.py
```diff
@@ -1,8 +1,33 @@
 from setuptools import setup
 
+
+"""
+This code is from the python documentation and is
+designed to read in the version number.
+See: https://packaging.python.org/en/latest/guides/single-sourcing-package-version/
+"""
+import codecs
+import os.path
```
JuliaKukulies, line 10: This goes back to a discussion @w-k-jones and me had on using `pathlib` instead of `os.path` (#113 ). One could also remove the `codecs` import, but then this would not be compatible with Python 2 anymore I guess. If we wanna be more Python 3, it could look like:
```

from pathlib import Path

def read(pkg_name):
    init_fname = Path(__file__).parent / pkg_name / '__init__.py'
    with open(init_fname, "r") as fp:
        return fp.read()

def get_version(pkg_name):
    for line in read(pkg_name).splitlines():
        if line.startswith("__version__"):
            delim = '"' if '"' in line else "'"
            return line.split(delim)[1]
    else:
        raise RuntimeError("Unable to find version string.")

PACKAGE_NAME = "tobac"

setup(
    name=PACKAGE_NAME,
    version=get_version(PACKAGE_NAME),
```

tobac/tests/test_import.py
```diff
@@ -4,3 +4,10 @@
 
 def test_dummy_function():
     assert 1 == 1
+
+
+def test_version():
+    """Test to make sure that we have a version number included.
+    If it's not, this should result in an error.
+    """
+    assert tobac.__version__ == tobac.__version__
```
JuliaKukulies, line 13: Note that we do not test the `get_version()` function in `setup.py` but only if we have defined `__version__` in `__init__.py` but I guess that is fine

JuliaKukulies (approved):
Thanks @freemansw1 for fixing this one quickly! I have a few thoughts, but I am also approving the PR as it is because it works fine. You can see if my suggestions are meaningful or if they do not really add any value :)

setup.py
JuliaKukulies: a suggestion to define a constant for the package name because "tobac" occurs now 3 times in setup.py (see l.30, 50)

tobac/tests/test_import.py
freemansw1: I'm not sure that we can test functions in setup.py? We also can't move these functions inside the library, because it isn't yet installed.

Integration testing is still performed in our CI workflow though, as we `pip install . ` in our Dockerfile.

setup.py
freemansw1: Good suggestion!

JuliaKukulies:
Thanks @freemansw1 for making these changes! Docker image failed because you forgot to update one line in my suggestion, but this should be fixed now

freemansw1:
Thanks @JuliaKukulies! I'll merge once @pjmarinescu or @kelcyno has a chance to review. -+

tobac/tests/test_import.py
JuliaKukulies: You are right and good point with the Docker image - actually the testing failed when you forgot to update one line in your last commit (see my comment below), which is great so we know this is covered in the workflow :)

freemansw1:
In the interest of moving this along, I've also requested @snilsn to review.

tobac/tests/test_import.py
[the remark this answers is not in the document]
freemansw1: Hm, good point. I don't know if test functions automatically `return` though? I will test that.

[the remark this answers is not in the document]
JuliaKukulies: Returning the version also works, but it does not seem so common to return values in tests - maybe because the only case in which the test fails is when the returned value is not defined? 

Alternatively `assert type(tobac.__version__) is str` ? That would also ensure the version attribute is the right type and not mistakenly `1.4` as a float or something. 

@snilsn Good point with the exception. I think so far we have only used `assert` statements that result in `AssertationError` when the tests fail, which is maybe good enough if the test function is properly described? Do you use  `pytest.raises(Exception)` also for that or do you use it only when you actually test if an exception is raised properly?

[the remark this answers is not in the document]
snilsn: After a bit of research I think the minimal be the way to do is just 

`version_number = tobac.__version__`

without a return. Or if we wanna be a little bit more elaborate and stick with  `assert`:

```
    try:
        version_number = tobac.__version__
    except Exception:
        assert False, 'no version number'
```

But we could also go with Julia's idea.

@JuliaKukulies Yes, I think `pytest.raises(Exception)` is mainly for testing if expected exceptions are raised correctly, but I'm really not an expert for testing. There could be clever ways to use this. Maybe we could make our tests more usefull by adding messages to our asserts, especially  if there is more than one within a function.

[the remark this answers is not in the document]
freemansw1: OK, I've now updated this code based on this discussion. Thanks much for the thoughts, @snilsn @JuliaKukulies

freemansw1:
Thanks @snilsn and @JuliaKukulies! I'll just wait for both of you to check over the new test, then I will merge.

snilsn (approved):

JuliaKukulies (approved):
Thanks @freemansw1, the new test looks good! Please go ahead and merge.

tobac/tests/test_import.py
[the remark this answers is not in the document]
JuliaKukulies: Thanks also for clarifying `pytest.raises(Exception)` !

freemansw1:
Thanks @snilsn and @JuliaKukulies !

Issue #158, opened 2022-07-09
Several `plotting.py` functions don't work - keep or ditch?


freemansw1:
As noted by @snilsn in #155, several of the functions in `plotting.py` do not currently work. His notebook is here: https://github.com/snilsn/tobac_functions/blob/master/plotting_functions.ipynb which contains a list and description of several of the broken functions.

As I have mentioned a few times before, I'm skeptical of the utility of most of the plotting functions in the first place; they come from the time when `tobac` was designed for use in just one paper rather than the more broadly used tool that it is today. I am inclined to immediately deprecate the broken functions, but I think this should be a more broad community discussion. The plotting functions will only become harder to maintain when we finish the Iris/Pandas -> xarray transition slated for v2.0, as many of them will need to be individually updated. 

For now, my suggestion is only to depreciate the broken functions (which I list below), but I do think we need to consider trimming these functions down more broadly as we start to think about the v2 transition. 

Known broken functions (all in `tobac.plotting`): 
- `plot_mask_cell_track_follow`
- `plot_mask_cell_individual_follow`
- `plot_mask_cell_track_static`
- `plot_mask_cell_individual_static`
- `plot_mask_cell_track_2D3Dstatic`
- `plot_mask_cell_track_3Dstatic`
- `plot_mask_cell_individual_track_3Dstatic`
- `plot_mask_cell_track_static_timeseries`

mheikenfeld:
Sorry to see that so many of the functions were left in here in a broken state... As Sean said, most of the code in the plotting functions has been developed for an initial paper using tobac, but then the code base of the rest of the package seems to have  moved on without updating the plotting functions accordingly (and without tests that has not been flagged up). I think I did actually move on to not really use these functions any more at some point and switched to more customized plotting functions functions myself, so I doubt anyone has really made explicit use of them at all in the last 2 years or so.

So my suggestion would be to maybe even deprecate the entire plotting bit v2 and rather cover that in a bunch of notebooks.

At the very least we should move all the functions out that outputs "png" or "pdf", seems like all of these are not really the right thing to keep as part of the package itself but rather something to make available to people as building blocks for vizualisation in the form of notebooks/documentation etc.

Maybe a couple of the functions could be useful in visualizing things on a very high level in tutorials for people starting out with tobac (such as plotting all tracks as a kind of Spaghetti plot and so on) or some usefull things could make it into the utils?

For whatever we keep, we should make sure we have it covered by simple tests, maybe based on the general test case we already have....

[...]

PR #183, opened 2022-10-13
Updates to codecov configuration

### LINKED ISSUE #159: Setting a lower threshold for Code Coverage to fail at
Right now, our codecov.io integration for CI fails when coverage drops at all. I'm not sure this is desired behavior, especially when coverage drops just a little bit. We should set some lower threshold for it to fail at. 

My initial suggestion is we allow a 1% point decrease in coverage before failing the CI, but I'm open to suggestions!

freemansw1:
Updates our codecov.io CI configuration to resolve #159 (and remove `setup.py`, which we can't test anyway) and update our GitHub Action to the newer version of the runner.

Note that I tested this on a clone on my account: https://github.com/freemansw1/tobac-codecov-test , https://app.codecov.io/gh/freemansw1/tobac-codecov-test as changing the CI configuration in a PR is hard to validate. It does appear to work, but I encourage the reviewers to examine those results, as this stuff is still new to me. 

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,22 @@
+# Codecov.io settings 
+
+coverage:
+  status:
+    project:
+      default:
+        # basic settings
+        # auto target means look at deltas from previous
+        target: auto
+        # Threshold we can drop coverage before failing CI
+        threshold: 1%
+        base: auto 
+        flags: 
+          - unit
+        paths: 
+          - "tobac"
+
+# what folders and files to ignore
+ignore:
+  - tobac/tests/.*
+  - setup.py
+
--- a/.github/workflows/codecov-CI.yml
+++ b/.github/workflows/codecov-CI.yml
@@ -26,7 +26,7 @@ jobs:
       run: mamba install --quiet --yes --file conda-requirements.txt coverage pytest-cov &&
            python -m coverage run -m pytest --cov=./ --cov-report=xml
     - name: Upload Coverage to Codecov
-      uses: codecov/codecov-action@v2
+      uses: codecov/codecov-action@v3
       with: 
         fail_ci_if_error: true
         flags: unittests


snilsn (approved):
Seems good to me @freemansw1 . `setup.py` is excluded in the test, the CI passes for this [commit ](https://app.codecov.io/gh/freemansw1/tobac-codecov-test/commit/c8ebef8283e645bdd46dcb27cddd885721aa0d56) with a drop in coverage under 1% and the notation matches the [Guide](https://docs.codecov.com/docs/codecovyml-reference).

The only thing i noticed is that in https://docs.codecov.com/docs/ignoring-paths the paths are in quotation marks, but that doesn't seem to make a difference.

freemansw1:
Thanks @snilsn and good point about the quotes. I've moved all paths to quotes.

codecov.yml
```diff
@@ -0,0 +1,22 @@
...
+        threshold: 1%
+        base: auto 
+        flags: 
+          - unit
+        paths: 
+          - "tobac"
+
+# what folders and files to ignore
+ignore:
+  - "tobac/tests/.*"
```
JuliaKukulies, line 20: Since we defined `omit = tobac/tests/*` in `.coveragerc`, do we actually have to define that again here or how do the `.coveragerc` and the configuration file for `codecov` work together?

JuliaKukulies: Would it make sense to also exclude lines that raise `NotImplementedError` and maybe the broken plotting functions discussed #158?

JuliaKukulies (commented):
I also had a look and it seems fine. Like @snilsn said, good to check that commits with a coverage decrease < 1% still passed! 

Two thoughts about which files to ignore, but I am approving the PR. Thanks for the updates @freemansw1 !

codecov.yml
freemansw1: You're right that we don't need to ignore/omit the files in both places for codecov.ci. I've kept this as-is though as if developers want to run coverage locally, `.coveragerc` is used, while `codecov.yml` is not. I will add `setup.py` to `.coveragerc` though to keep things consistent. 

I don't believe that we can exclude specific lines of functions. I'm working on a PR though that addresses #158 and will add (limited) tests for those, though.

freemansw1:
Thanks @JuliaKukulies ! Since I've made a few changes, @snilsn are you still happy with me merging this?

snilsn:
Sure @freemansw1! Please go ahead

freemansw1:
Thanks!

PR #190, opened 2022-10-23
Split up utils module similar to `v2.0-dev`

freemansw1:
This should address #122 and bring our `utils` module closer to what we have in `v2.0-dev`. This should pair nicely with #179 and is another (small) step toward bridging the gap between `v2.0-dev` and v1.x. 

* [x] Have you followed our guidelines in CONTRIBUTING.md? 
* [x] Have you self-reviewed your code and corrected any misspellings? 
* [ ] Have you written documentation that is easy to understand?
* [x] Have you written descriptive commit messages? 
* [ ] Have you added NumPy docstrings for newly added functions? 
* [x] Have you formatted your code using black? 
* [ ] If you have introduced a new functionality, have you added adequate unit tests?
* [x] Have all tests passed in your local clone? 
* [ ] If you have introduced a new functionality, have you added an example notebook?
* [x] Have you kept your pull request small and limited so that it is easy to review? 
* [x] Have the newest changes from this branch been merged?

Issue #122, opened 2022-04-17
Split `utils` module into internal utilities and external utilities


freemansw1:
We should split the `utils.py` module into an internal utilities and external utilities module to save ourselves (and our users) some headaches down the road. 

### Proposal
Keep all external utilities in `utils.py` to preserve backwards compatibility, move all internal utilities to something like `internal_utils.py`. 

#### Proposed division:
**External Utilities**
- `column_mask_from2D`
- `mask_cube_cell`
- `mask_cube_all`
- `mask_cube_untracked`
- `mask_cube`
- `mask_cell`
- `mask_cell_surface`
- `mask_cell_columns`
- `mask_cube_features`
- `mask_features`
- `mask_features_surface`
- `mask_all_surface`
- `get_spacings`

**Internal Utilities**
- `add_coordinates`
- `get_bounding_box`
- `get_label_props_in_dict`
- `get_indices_of_labels_from_reg_prop_dict`

JuliaKukulies:
I think that this subdivision would make sense. Maybe more than the division into `mask.py` and `general.py` as in the subpackage `utils` on `v2.0-dev`  ? If the plan is to keep `mask.py` and `general.py` in `v2.0-dev` , then it would maybe be better to split `utils.py` in the same way it is done there. But I think your proposal sounds good and could be moved to `v2.0-dev`  as well.

freemansw1:
Great point @JuliaKukulies! When dealing with this particular headache, I hadn't thought to look how `v2.0-dev` handled things. I think we can actually go ahead and move `utils` into its own submodule without breaking anyone's existing code and start to bring things closer to the way they should be in `v2.0-dev`. I'll look into this and put in a PR whenever I get time.

JuliaKukulies:
@freemansw1 I wanted to start working on a PR to bring the spectral filtering into `dev` so it will be ready for v1.5. Do you think I should wait until we have moved `utils` into a submodule? In `v2.0-dev`, I put `spectral_filtering` in `general.py`, but I am happy to move it somewhere else if you think it fits better in `feature_detection.py` itself for example.

freemansw1:
I am inclined, for now, to just keep everything in `utils.py`? That's what we've done with the 3D/PBC work that should be in a PR today or tomorrow. After both of those are merged (and splits/mergers), we can separate things out.

JuliaKukulies:
Ok, sounds good!

freemansw1:
Resolved with #191

PR #136, opened 2022-06-06
Merge Split algorithm added

kelcyno:
The addition in this pull request is a post processing algorithm for the tobac tracking. This file contains the merge_split function to address merging/splitting cells in the tobac track, as well as a quick compression function, and a function to standardize the track dataset in order to have all tracking, feature, and segmentation mask datasets in one xarray dataset. 
An example of these functions is given below:  

d = merge_split(Track)
ds = standardize_track_dataset(Track, refl_mask, data['ProjectionCoordinateSystem'])
 both_ds = xr.merge([ds, d],compat ='override')
 both_ds = compress_all(both_ds)
 both_ds.to_netcdf(os.path.join(savedir,'Track_features_merges.nc'))

PR #191, opened 2022-10-24
Split up utils module similar to `v2.0-dev`

freemansw1:
This should address https://github.com/tobac-project/tobac/issues/122 and bring our utils module closer to what we have in `v2.0-dev`. This should pair nicely with https://github.com/tobac-project/tobac/pull/179 and is another (small) step toward bridging the gap between `v2.0-dev` and `v1.x`.

Note that the way that I've done this should preserve all existing code. We will probably want to depreciate public use of some of these functions (e.g., why would a user want to call `add_coordinates`), but that is not this PR. 

(this is a re-do of #190)


freemansw1:
I should note that the only actual code change that should be here is the deletion of the two commented out functions in `uitls.py`. I didn't see much of a point to keeping those around.

snilsn:
I haven't gone through this in detail yet, but one thing I wondered:

There is now a separate `__init__.py` for tobac.utils, but there are still imports from tobac.utils in the original `__init__.py`.

That means I could access the same function in three different ways, e. g.

```
import tobac
tobac.mask_features()
tobac.utils.mask_features()
tobac.utils.mask.mask_features()
```
would all call the same function. Is that the way it's meant to be? I can imagine that this could be confusing.

JuliaKukulies:
@snilsn I think having a separate `__init__.py` file is not a bad idea for submodules like the new `utils` that contain multiple files and functions and is also the module that is the most likely to rapidly grow. I see your point that it is a bit confusing to have three different ways of calling a function and converting `utils` into a package itself would also have the advantage of just calling `utils` in `tobac.__init__.py` and access all functions independent in which submodule of `utils` they are. I could, however, imagine that @freemansw1 kept the original imports to not break anybody's code when they usually call the functions with `tobac.mask_features()` etc. Is that right? 


@freemansw1 Agreed with the removal of the functions that were commented out.

freemansw1:
Yes, what @JuliaKukulies says is exactly the reason that I structured the code the way that I did. Any functions that were exposed via `import tobac` are still exposed to the user through this mechanism, and this should only break code if people tried to `import tobac.utils` and call a function not exposed through `import tobac` directly from there. The functions that would apply to are largely functions that I doubt users are calling anyway (e.g., `get_label_props_in_dict`).

I'm not opposed to deprecating that way of calling functions, and encouraging users to switch to importing `tobac.utils` instead, but I think that's a different PR (and I don't immediately know how to do that).

That said, this doesn't necessarily avoid breaking everyone's code. If a user is directly using a function from `tobac.utils` that is not exposed now, their code would break without a deprecation warning. As far as I can tell, the functions that this would apply to are:

- `tobac.utils.get_label_props_in_dict`
- `tobac.utils.get_indices_of_labels_from_reg_prop_dict`

Maybe it makes sense to import those functions from `utils/__init__` to really avoid breaking anyone's code?

snilsn:
Thank for this carification @JuliaKukulies and @freemansw1, that makes a lot of sense. I agree that deprecating this way of calling functions is probably not necessary and certainly not urgent.

tobac/utils/mask.py
```diff
@@ -0,0 +1,408 @@
...
+    """
+
+    feature_ids = track.loc[track["cell"] == cell, "feature"].values
+    mask_i_surface = mask_features_surface(
+        mask, feature_ids, masked=masked, z_coord=z_coord
+    )
+    return mask_i_surface
+
+
+def mask_cell_columns(mask, cell, track, masked=False, z_coord="model_level_number"):
```
snilsn, line 222: I would suggest deleting this function as well, since it depends on a function that is now deleted.

snilsn (approved):
Well done @freemansw1, one more step towards v1 and 2 compatibility! I have just one suggestion.

It's interesting to see how few of these utils are actually used internally, I suspect that many of the masking functions never have been in use.

JuliaKukulies (approved):
Nice work @freemansw1! It makes sense how you divided the functions, especially since you also separated out the functions for masking. And I think you thought of everything to make this work, so I do not see any need for changes here.

tobac/utils/mask.py
freemansw1: Good point, @snilsn . I've deleted it in my latest commit.

freemansw1:
After thinking about it some, I've added the two functions mentioned above to `tobac.utils` such that no code (should) be broken. 

I've also addressed @snilsn's comment on deleting the not implemented function.

JuliaKukulies:
I have realized that changing this structure also affects the auto-generated API reference and does not show any of the new submodules in `tobac.utils`,  see https://tobac--191.org.readthedocs.build/en/191/tobac.html#module-tobac.segmentation 

I am not sure if this has something to do with the default options that we used for the napolean extension in `doc/conf.py` that submodules of submodules are not shown, or could it be due to the same reason why `merge_split.py` is not shown (see #136 )? Maybe a better way to control this is to create an  [api_reference.rst](https://brandons-sphinx-tutorial.readthedocs.io/en/latest/writing-api.html) file. @freemansw1 do you have any thoughts?

freemansw1:
Good catch, @JuliaKukulies . I've added the `tobac.utils.general` and `tobac.utils.mask` modules to the API reference, but not the `tobac.utils.internal`, on the theory that internal utils shouldn't be used by users anyway. That said, we don't explicitly label these with `_` to mark them as internal, so I'm happy to add them to the documentation if we think that's what is best.

freemansw1:
OK, with #179 merged, I'm re-requesting reviews and will merge after those go through.

JuliaKukulies (approved):
The update looks good @freemansw1 ! And it makes sense to me that you added the new functions for conversion to`internal.py` instead of `convert.py`

snilsn (approved):
Looks great @freemansw1 and I agree on adding the converters to the internal utils.

Issue #185, opened 2022-10-20
Drop support for Python <3.7


freemansw1:
Conda-forge is about to drop support for Python 3.7 (https://conda-forge.org/docs/user/announcements.html#dropping-python-3-7) and they have already dropped support for Python 3.6 and below. With that, I think it would be in our best interest to go ahead and require at least Python 3.7 starting with _tobac_ v1.4.0. This will also help us move the code base forward and gives us the option to include features such as type annotations (see #75).

Does anyone see issues with this before I create a PR?

JuliaKukulies:
That sounds like a good plan @freemansw1 !

freemansw1:
We've now added the deprecation warning on import (#193), so the next step is to update the meta.yml file, which I'll do on release of 1.4.

PR #186, opened 2022-10-21
New function to combine feature dataframes

freemansw1:
This should address #17 (!) and provides a function for combining multiple, independently identified feature dataframes into a single dataframe appropriate for tracking. 

* [x] Have you followed our guidelines in CONTRIBUTING.md? 
* [x] Have you self-reviewed your code and corrected any misspellings? 
* [x] Have you written documentation that is easy to understand?
* [x] Have you written descriptive commit messages? 
* [x] Have you added NumPy docstrings for newly added functions? 
* [x] Have you formatted your code using black? 
* [x] If you have introduced a new functionality, have you added adequate unit tests?
* [x] Have all tests passed in your local clone? 
* [x] If you have introduced a new functionality, have you added an example notebook?
* [x] Have you kept your pull request small and limited so that it is easy to review? 
* [x] Have the newest changes from this branch been merged?

Issue #176, opened 2022-09-09
Add tests for other python versions


freemansw1:
Currently, we only run tests on the latest versions of Python and our dependent libraries. We should probably expand this to test multiple python versions. This should be doable through GitHub Actions, similar to how Metpy has done it. 

I'm inclined to have us test the latest minor release for every Python version that we support (we are >=3.6 now). It might also be a good idea to test multiple versions of our libraries, but I see a less clear path forward on that. 

@mgrover1 given your immense experience here, I'd be interested in hearing your opinion on this!

JuliaKukulies:
That sounds like a very good idea @freemansw1 and as you say pretty straightforward to add different python versions.

I have participated in another coding project where we also have done this: 
https://github.com/SEE-GEO/pansat/blob/main/.github/workflows/install_and_test.yml 
(similar to Metpy). 


On the same note, we could also test on operating systems other than Ubuntu, or have we deliberately chosen not to do so?

freemansw1:
With us accidentally breaking compatibility in #175, I've tagged this to be done in v1.5, which is when we drop support for <3.7. 

> On the same note, we could also test on operating systems other than Ubuntu, or have we deliberately chosen not to do so?

No particular reason, other than the Dockerfile was a copy/paste of something else I had done. We can add other linux OSes to our test suite, but given that we don't have any compiled code, it may be redundant.

mgrover1:
Happy to help with this effort!

w-k-jones:
Resolved with #353

PR #193, opened 2022-10-24
Deprecate Python 3.6 and lower support

freemansw1:
Resolves the first half of #185. Rather than dropping support for <3.7, I'm proposing that we only deprecate it in the code. We will still tell conda-forge that we require >=3.7, but this allows users who must remain on old versions to stick around for now. We will drop support with v1.5, as I'd really like to start adding typing annotations. 

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ __pycache__
 .vscode
 htmlcov
 .coverage
+build
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ Detailed documentation of the package can be found at https://tobac.readthedocs.
 
 Installation
 ------------
-tobac now works for both Python 3 and Python 2 installations.
+tobac requires Python 3, and support for Python versions before 3.7 (i.e., 3.6 and lower) is deprecated and will be removed in tobac version 1.5.
 
 The easiest way is to install the most recent version of tobac via conda and the conda-forge channel:
 ```
--- a/setup.py
+++ b/setup.py
@@ -1,8 +1,5 @@
 from setuptools import setup
 
-import sys
-import warnings
-
 """
 This code is from the python documentation and is
 designed to read in the version number.
@@ -24,14 +21,6 @@ def get_version(pkg_name):
             return line.split(delim)[1]
     else:
         raise RuntimeError("Unable to find version string.")
-
-
-if sys.version_info < (3, 7):
-    warnings.warn(
-        "Support for Python versions less than 3.7 is deprecated. "
-        "Version 1.5 of tobac will require Python 3.7 or later.",
-        DeprecationWarning,
-    )
 
 
 PACKAGE_NAME = "tobac"
--- a/setup.py
+++ b/setup.py
@@ -1,5 +1,6 @@
 from setuptools import setup
 
+import sys
 
 """
 This code is from the python documentation and is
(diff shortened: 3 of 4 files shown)


JuliaKukulies (changes requested):
I agree with your comment on type annotations for `v1.5` and I like your suggestion to first deprecate older python versions instead of directly dropping the support @freemansw1! 

However, I am not sure what the best way to do this is. I tested to install `tobac` with Python 3.6 and even though your code works perfectly fine when run outside of `setup.py`, the `DeprecationWarning` does neither show up when installing `tobac` from the local clone (`pip install tobac/`) nor when `tobac` is installed using `conda`. 

For Python versions > 3.7, the `DeprecationWarning` shows up when using the verbose flag: `pip install -v tobac/`, but this does not really help us, because obviously the warning is only raised for older versions and even when it shows up it is a bit difficult to read with all the other output `setup.py` gives.

I am trying to find out what others have done. Maybe just state in the `README`? 

Also, do I understand it correctly that you propose to already now require `>3.7` for `conda `so that users who install `tobac v1.4.0` via `conda` would need to do so with Python >3.7 while users who need to remain with older Python versions need to install `tobac v1.4.0` from the github repo?

JuliaKukulies:
OK, I tried some things and what worked is to use a print statement and put it at the end of `setup.py` and add some blank lines so that it sticks out from the other output. This does, however, also only work when installing `tobac` from source with `python -v install .` My suggestion would be to just add it here but also write the same to the `README` because most users probably do not use the `verbose flag`. 

```python 

if sys.version_info[0] < 3.7:
    warning = """ \n\n 
    Support for Python versions less than 3.7 is deprecated. 
    Version 1.5 of tobac will require Python 3.7 or later.
   Python {py} detected. \n\n
    """.format(py='.'.join(str(v) for v in sys.version_info[:3]))

    print(warning)  
```

freemansw1:
> Also, do I understand it correctly that you propose to already now require >3.7 for conda so that users who install tobac v1.4.0 via conda would need to do so with Python >3.7 while users who need to remain with older Python versions need to install tobac v1.4.0 from the github repo?

Yes, that's my proposal. I think that will happen anyway, even without our explicit doing, due to the fact that conda-forge doesn't support 3.7 or lower anymore. If a user needs to be on Python <3.6, they can still get `tobac` for now, but it's now 6 years old, so I am not opposed to nudging users to upgrade sooner rather than later.

JuliaKukulies (approved):
Thanks for these changes @freemansw1! That was a good idea to switch the warning to import instead. I also tested it one more time in a python 3.6 environment and it works fine.

JuliaKukulies:
And good point with #176. I am almost inclined to say that we are OK with it due to the reasons that you mention (better sooner than later nudge users to upgrade and users installing `tobac` from now with  `conda-forge` with python<3.4 will get `tobac1.2` anyhow). I would think that a smooth transition from Python3.6 is more important because this probably concerns much more people, but if others have a different opinion, we can definitely work out a PR that still supports Python 2 for the new release. In #176 this only concerns the `pathlib` module, right? Note that we also used it in the updated examples (#186).

snilsn (approved):
Good work, @freemansw1, I didn't find any issues. Giving a warning on import is probably the best way to do this, since it's a more frequent reminder.

freemansw1:
Thanks! 

Agreed with you @JuliaKukulies . If users are still on Python 2.7 or <3.4, they can use tobac v1.3.3. If they're on those python versions, they must be installing from GitHub anyway as conda-forge doesn't support those versions.

Issue #196, opened 2022-10-26
Add scikit-learn as a requirement?


freemansw1:
Given that it's used in #127 and could be used eventually in #136, I think we should add `scikit-learn` to our dependencies. It's a really common dependency for other libraries in this space, and many of the items in there are exceedingly useful, particularly their very fast code for calculating distances. 

The downside of this is, of course, adding another dependency. Scikit-learn is a big piece of code, and isn't as "standard" as some of our other dependencies (e.g., numpy, pandas, etc.). 

The other path forward, given that we use it in #127, would be to make it an optional dependency and required for specific features to be used, similar to how we are planning on handling mergers and splits.

freemansw1:
I'm going to put this on our agenda to discuss at the next dev meeting next week.

freemansw1:
After discussion at the dev meeting last week, scikit-learn will be added as a dependency in v1.5.0.

freemansw1:
Resolved with #204

Issue #192, opened 2022-10-24
Put *tobac* on PyPI


freemansw1:
We should probably put *tobac* on PyPI. This isn't urgent and it does make our release process more cumbersome, but it's important for non-conda users. The tutorial is here: https://packaging.python.org/en/latest/tutorials/packaging-projects/

JuliaKukulies:
Good idea @freemansw1 and should probably not cause any problems given that all dependencies in `requirements.txt` exist on PyPI and given that we do not have any external dependencies. 

I am not sure how much more cumbersome this will make our release process though. I have no experience with it but since there are many packages (e.g. `MetPy`) that are published on both the Anaconda cloud and PyPI, we can double check how they do it. I understand from [this blog post](https://levelup.gitconnected.com/publishing-your-python-package-on-conda-and-conda-forge-309a405740cf) that one can specify the location on PyPI as a source in the conda recipe, so that the package on conda-forge is automatically updated when the package is updated on PyPI. Not sure if it also works the other way around?

PR #204, opened 2022-11-17
Update requirements to require scikit-learn and purge old requirements

### LINKED ISSUE #119: Pytables dependency removal
Source code inspection for the conda-forge feedstock found that pytables is not referenced anywhere and could be removed as a dependency. Is this true, and if so should it be removed from both the conda-forge and local requirements?

freemansw1:
Resolves #119 and #196. 

**Specific actions**
* Removes `pytables` and `cf-units` from our dependencies (both are used by our dependencies, but not by us directly)
* Renames `conda-requirements.txt` to `requirements.txt` as that seems to be the standard (I cannot find reference to `conda-requirements.txt` anywhere)
* Adds requirements to `setup.py` such that `pip install .` installs dependencies as well. 

Standard questions


freemansw1:
Note that the requirements in `setup.py` do not match those in `requirements.txt` because cartopy builds through `pip` appear to be broken, and our `black` runner depends on `pip install .` working.

snilsn (changes requested):
Good job @freemansw1, makes sense to me. There is a minor issue that I have found.

The README.md still references the old name of requirements.txt

`conda install -c conda-forge --yes --file conda-requirements.txt`

snilsn (approved):
Great! Approving this now

setup.py
```diff
@@ -49,6 +49,15 @@ def get_version(pkg_name):
...
+    install_requires=[
+        "numpy",
+        "scipy",
+        "scikit-image",
+        "scikit-learn",
+        "pandas",
+        "matplotlib",
+        "xarray",
+        "trackpy",
+    ],
```
JuliaKukulies, line 61: What about `iris` and `cartopy` here? 

And should we maybe add `pytest` as `test_requires` ?

JuliaKukulies (commented):
Great update @freemansw1!

setup.py
freemansw1: See above comment about iris and cartopy. I'm not sure that we can do much until/unless cartopy fixes their pip builds. 

Good catch with `test_requires`, I'll add that shortly.

JuliaKukulies:
All right, approved! 

And sorry that I missed your very clear comment on cartopy and iris! You are probably right that we can not do much about that.

freemansw1:
> All right, approved!
> 
> And sorry that I missed your very clear comment on cartopy and iris! You are probably right that we can not do much about that.

No worries! It is a bit odd, and I do worry that this will cause us headaches when putting things onto PyPI (#192), but it's the solution that we have for now.

freemansw1:
I had to resolve a merge conflict with #191 merged in. @JuliaKukulies or @snilsn can you give this a quick look-over again and then I will merge?

snilsn:
Looks good and ready to be merged @freemansw1!

freemansw1:
Thanks!

PR #223, opened 2023-01-12
Resolves #212

### LINKED ISSUE #212: Raise error if no search range parameter is specified
We should raise a specific error if neither `v_max`, `d_max` nor `d_min` is provided in `linking_trackpy`.

lettlini:
This pull request should resolve #212.

A simple check has been added to correctly handle the case of neither `v_max, d_min, d_max` being passed to `linking_trackpy`, i.e. a `ValueError` is raised if all three arguments are missing. A corresponding test case has also been added.

### DIFF
--- a/tobac/tracking.py
+++ b/tobac/tracking.py
@@ -180,7 +180,7 @@ def linking_trackpy(
 
     if (v_max is None) and (d_min is None) and (d_max is None):
         raise ValueError(
-            "Neither one of d_min, d_max, v_max has been provided. Exactly one of these arguments must be specified."
+            "Neither d_max nor v_max has been provided. Either one of these arguments must be specified."
         )
 
     # calculate search range based on timestep and grid spacing
--- a/tobac/tracking.py
+++ b/tobac/tracking.py
@@ -179,7 +179,9 @@ def linking_trackpy(
     #    from .utils import add_coordinates
 
     if (v_max is None) and (d_min is None) and (d_max is None):
-        raise ValueError("Neither one of d_min, d_max, v_max has been provided. Exactly one of these arguments must be specified.")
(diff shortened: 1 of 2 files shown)


JuliaKukulies (approved):
This looks great,  @lettlini!  No objections to merging the code changes as they are. Thank you for fixing this one

tobac/tracking.py
```diff
@@ -178,6 +178,11 @@ def linking_trackpy(
     #    from trackpy import filter_stubs
     #    from .utils import add_coordinates
 
+    if (v_max is None) and (d_min is None) and (d_max is None):
```
snilsn, line 181: Now that I look again at the explanation of `d_min` in the docstring, I think we should exclude it here, since its original purpose is not to define a search range (The current implementation, however, only mimics `d_max`).

Since the parameter will eventually be deprecated, it is probably better not to promote its use in an error.

snilsn (commented):
Good job Kolya @lettlini. I have only one thought to discuss, but this is due to my formulation of #212

tobac/tracking.py
freemansw1: I agree with your thoughts here, Nils. I'll wait to review until this is changed. I'd say that we should leave`d_min` in the if statement here, but remove it in the error message.

lettlini: I think you are both absolutely right. I have updated the error message accordingly.

snilsn (approved):

freemansw1:
@lettlini since Nils has approved, I'm happy to merge this; are you still happy for it to be included?

lettlini:
@freemansw1 Sorry for the late reply. Yes I am still happy for this PR to be merged

Issue #98, opened 2022-03-25
How to link segmentations based on lon and lat coordinates?


zxdawn:
I see @JuliaKukulies has added the lon/lat support (#65), is it possible to link the 2D segmentations with different shapes? Let's assume we have two segmentations: seg1 (y:200, x: 100) and seg2 (y: 190, x: 120). Both of them have 2D lon and lat coordinates. If we can utilize the lon/lat coordinates with segmentation info together. that would be awesome!

freemansw1:
We have some code here that we've been working on to do basically that. It's close but not yet ready for a PR. Hopefully, we can get this in for v1.4 or v1.5.

w-k-jones:
Transferring detected features and segmentations across different grids is definitely a feature on the radar for tobac v2.0. This sounds like a subset of that situation, where it's only applied to two lat/lon grids (rather than grids on different coordinates). A couple of questions to see how complex this is:

- Do the lat/lon grids share the same points and grid spacing as each other. I.e. are the offset from each other but overlapping?
- Do they occupy the same spatial extent, but the grid spacing is different?

There are two approaches to this I've considered. The first, more direct approach is simply to regrid from one grid to the other. This is pretty straightforward, but more computationally intensive and slightly less satisfactory, particularly if you want to preserve resolution. The second approach is to turn each segmentation mask into a set of geometric objects that contain each of the segmented features, which could then be mapped to any other grid without having the regrid the data on the second grid itself

zxdawn:
@w-k-jones Thanks for the idea. For polar-orbiting satellite data, the lon/lat, spacing, and spatial extent are all different. Here's an example:

![image](https://user-images.githubusercontent.com/30388627/160241362-71c031ba-8f11-452d-86ab-539d6965f488.png)
![image](https://user-images.githubusercontent.com/30388627/160241344-34ebdef5-bb86-49d6-a7b7-b1fe0706fa97.png)

I guess I can use geopandas to convert segmentations into geometric objects, but how to map it to another grid?

w-k-jones:
Here's a quick mock-up of what I'm thinking:

<img width="807" alt="remap_mask_example" src="https://user-images.githubusercontent.com/22833757/160249086-d620c06d-ed8d-4160-a245-f7eea0d30a20.png">

[...]

PR #242, opened 2023-02-09
Adding ability to track on one dataset, segment on another

freemansw1:
This is new functionality that I first mentioned in #98 to transform feature points into a new grid to allow for segmenting on a different grid from the one you are tracking on. 

Note the change to the time comparisons in the segmentation code; iris times gave me a pretty good headache, but I think the solution is universal until we switch over to xarray. 

* [x] If you have introduced a new functionality, have you added an example notebook? (https://github.com/tobac-project/tobac-tutorials/pull/28)


tobac/utils/general.py
```diff
@@ -553,3 +554,129 @@ def combine_tobac_feats(list_of_feats, preserve_old_feat_nums=None):
...
+
+
+@internal_utils.irispandas_to_xarray
+def transform_feature_points(
+    features,
+    new_dataset,
+    latitude_name="auto",
+    longitude_name="auto",
+    max_time_away=datetime.timedelta(minutes=1),
+    max_space_away=20000,
```
w-k-jones, line 566: I'd prefer these set to `None` (or `np.inf`) by default rather than fixed values. The fixed values could lead to unexpected behaviour for some users (e.g. if they have data with low spatial/temporal resolution)

tobac/utils/general.py
```diff
@@ -553,3 +554,129 @@ def combine_tobac_feats(list_of_feats, preserve_old_feat_nums=None):
...
+    # there is almost certainly room for speedup in here.
+    for index in features["index"]:
+        dist, closest_pt = ll_tree.query(
+            [
+                [
+                    np.deg2rad(features["latitude"][index]),
+                    np.deg2rad(features["longitude"][index]),
+                ]
+            ]
+        )
```
w-k-jones, line 646: Can we pass all of the feature locations to ll_tree in one call?

tobac/utils/general.py
```diff
@@ -553,3 +554,129 @@ def combine_tobac_feats(list_of_feats, preserve_old_feat_nums=None):
...
+
+    # find where distances are too large and drop them.
+    new_lat = lat_vals_new[new_h1, new_h2]
+    new_lon = lon_vals_new[new_h1, new_h2]
+    dist_apart = (
+        tb_analysis.haversine(
+            new_lat, new_lon, features[lat_coord], features[lon_coord]
+        )
+        * 1000.0
+    )
```
w-k-jones, line 665: Does this differ from the distance returned from querying the ll_tree? If not, would it be more efficient to store that value and use it instead?

tobac/utils/general.py
```diff
@@ -553,3 +554,129 @@ def combine_tobac_feats(list_of_feats, preserve_old_feat_nums=None):
...
+    ret_features = ret_features.where(dist_apart < max_space_away, drop=True)
```
w-k-jones, line 666: This drops features silently, would it be good to add a warning or logging for when features are dropped due to exceeding the distance threshold so as to make this behaviour clear?

tobac/utils/general.py
```diff
@@ -553,3 +554,129 @@ def combine_tobac_feats(list_of_feats, preserve_old_feat_nums=None):
...
+
+    # force times to match, where appropriate.
+    if "time" in new_dataset.coords:
+        # this is necessary due to the iris/xarray/pandas weirdness that we have.
```
w-k-jones, line 670: When we convert to xarray internally throughout could we change this to using `.sel(..., method="nearest")`?

tobac/utils/general.py
```diff
@@ -553,3 +554,129 @@ def combine_tobac_feats(list_of_feats, preserve_old_feat_nums=None):
...
+        old_feat_times = ret_features["time"].astype("datetime64[s]")
+        closest_times = np.min(np.abs(old_feat_times - new_dataset["time"]), axis=1)
+        closest_time_locs = np.abs(old_feat_times - new_dataset["time"]).argmin(axis=1)
+        # force to seconds to deal with iris not accepting ms
+        ret_features["time"] = new_dataset["time"][closest_time_locs].astype(
+            "datetime64[s]"
+        )
+        ret_features = ret_features.where(
+            closest_times < np.timedelta64(max_time_away), drop=True
+        )
```
w-k-jones, line 680: Like with the distance filter, we should probably make the user aware that this is happening

tobac/utils/internal.py
```diff
@@ -662,3 +662,46 @@ def find_hdim_axes_3D_iris(field_in, vertical_coord=None, vertical_axis=None):
...
+    -------
+    lat_name, lon_name: tuple(str)
+        the detected names of the latitude and longitude coordinates. If
+        coordinate is not detected, returns None for that coordinate.
+
+    """
+    out_lat = None
+    out_lon = None
+    test_lat_names = ["lat", "latitude"]
+    test_lon_names = ["lon", "long", "longitude"]
```
w-k-jones, line 690: At some point we should move all of the alternative coord names to a separate .json file or similar so that they can be managed more easily

Alternatively, does cf-units handle alternative coord names?

tobac/utils/internal.py
```diff
@@ -662,3 +662,46 @@ def find_hdim_axes_3D_iris(field_in, vertical_coord=None, vertical_axis=None):
...
+                out_lat = test_lat_name
+                break
+    if longitude_name != "auto":
+        if longitude_name in in_dataset.coords:
+            out_lon = longitude_name
+    else:
+        for test_lon_name in test_lon_names:
+            if test_lon_name in in_dataset.coords:
+                out_lon = test_lon_name
+                break
```
w-k-jones, line 706: Is this case sensitive? Might be worth looking into whether this could be made case insensitive in future

w-k-jones (commented):
Looks good at first glance, my main concerns are regarding unexpected behaviour in `transform_feature_points`. Otherwise most of my comments are things that are probably best left for when we refactor everything for xarray in version 2.

I'll check out the branch and try it with some different data later and let you know if any issues arise

tobac/utils/general.py
```diff
@@ -553,3 +554,129 @@ def combine_tobac_feats(list_of_feats, preserve_old_feat_nums=None):
...
+        * 1000.0
+    )
+    ret_features = ret_features.where(dist_apart < max_space_away, drop=True)
+
+    # force times to match, where appropriate.
+    if "time" in new_dataset.coords:
+        # this is necessary due to the iris/xarray/pandas weirdness that we have.
+        old_feat_times = ret_features["time"].astype("datetime64[s]")
+        closest_times = np.min(np.abs(old_feat_times - new_dataset["time"]), axis=1)
+        closest_time_locs = np.abs(old_feat_times - new_dataset["time"]).argmin(axis=1)
```
JuliaKukulies, line 673: This can easily result in memory issues when the total number of features and the total number of timesteps in the new dataset are large. However, iterating over each feature will be also quite ineffective, so we might have to accept that ?

JuliaKukulies (changes requested):
Great addition @freemansw1 ! Looks overall good..just a couple of small comments and one issue where I am not sure if we have to address this in this PR (see comment about `standard_name` used in iris conversion)

tobac/utils/general.py
```diff
@@ -553,3 +554,129 @@ def combine_tobac_feats(list_of_feats, preserve_old_feat_nums=None):
...
+    max_time_away=datetime.timedelta(minutes=1),
+    max_space_away=20000,
+):
+    """Function to transform input feature dataset horizontal grid points to a different grid.
+    The typical use case for this function is to transform detected features to perform
+    segmentation on a different grid.
+
+    The existing feature dataset must have some latitude/longitude coordinates associated
+    with each feature, and the new_dataset must have latitude/longitude available with
+    the same name.
```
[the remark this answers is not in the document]
w-k-jones: I've encountered this issue as well. Ideally we should modify the iris cube so that it has the same coordinate names as the xarray dataset (i.e. the short name rather than standard name). Mostly this is so that there is continuity in the coordinate naming across conversions, but there are also some issues that can arise with multiple coordinates having the same standard name

I think that this is beyond the scope of this PR, but something we should address for the 1.6 release

tobac/segmentation.py
```diff
@@ -588,7 +589,7 @@ def segmentation(
 
     for i, field_i in enumerate(field_time):
         time_i = field_i.coord("time").units.num2date(field_i.coord("time").points[0])
-        features_i = features.loc[features["time"] == time_i]
+        features_i = features.loc[features["time"] == np.datetime64(time_i)]
```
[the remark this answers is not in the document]
freemansw1: Yes, it fixes cases where the Pandas -> Xarray -> Iris conversion of times produces two functionally equivalent times from `cf-units` but do not evaluate as equal. This will probably not be necessary with the conversion to xarray everywhere. I believe that it shouldn't change behavior otherwise.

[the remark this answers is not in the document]
w-k-jones: We should make a list of all the iris -> xarray oddities to make sure we can cover them with conversion decorators for the v1.6 release

tobac/utils/general.py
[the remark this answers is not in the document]
JuliaKukulies: I agree, that makes sense!

tobac/utils/general.py
JuliaKukulies: Good idea! Maybe we can make sure then that `time` is defined as a dimension and not just a coordinate as I find is quite often the case in poorly post-processed model data ;)

tobac/utils/general.py
freemansw1: I've set these to a default value of 0 and 0.

tobac/utils/general.py
freemansw1: I've now refactored this section to consolidate these queries to one call.

tobac/utils/general.py
freemansw1: I've fixed this, and avoid the use of our `haversine`

tobac/utils/general.py
freemansw1: I am disinclined from adding warning messages as they can get quite expensive for large datasets. It would be better to have a flag in the function to keep (but disable in some way) the output features, but we would need to add the ability in segmentation to detect these bad features and not attempt to segment on them. I'm hesitant to dig into that rabbit hole for now.

tobac/utils/general.py
freemansw1: See response above.

tobac/utils/general.py
freemansw1: Yes, agreed all around!

tobac/utils/internal.py
freemansw1: Agreed on the eventual need for a refactor around coord names, once xarray is in. 

Once we banish Iris, I'm inclined to also banish cf-units. There are so many cases where units (especially when trying to do things outside of earth science) don't cleanly match up with cf-units.

tobac/utils/internal.py
freemansw1: I've fixed case sensitivity to "auto" but have left it for the others, given that one could theoretically have both a `LAT` and `lat` dimension (although that would be silly).

doc/transform_segmentation.rst
```diff
@@ -0,0 +1,9 @@
+Track on one dataset, segment on another
```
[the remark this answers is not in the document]
freemansw1: Good catch.

tobac/utils/internal.py
```diff
@@ -662,3 +662,46 @@ def find_hdim_axes_3D_iris(field_in, vertical_coord=None, vertical_axis=None):
...
+
+    Returns
+    -------
+    lat_name, lon_name: tuple(str)
+        the detected names of the latitude and longitude coordinates. If
+        coordinate is not detected, returns None for that coordinate.
+
+    """
+    out_lat = None
+    out_lon = None
```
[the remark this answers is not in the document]
freemansw1: I am disinclined from doing so at the moment, because different functions could still proceed without lat/lon being present. One could put in a try/catch block, but this feels cleaner to me. Happy to switch, though.

doc/transform_segmentation.rst
```diff
@@ -0,0 +1,9 @@
...
+----------------------------------------
+
+*tobac* also has the capability to combine datasets through :doc:`segmentation`, which includes the ability to track on one dataset (e.g., gridded radar data) and run segmentation on a different dataset *on a different grid* (e.g., satellite data).
+
+        .. image:: images/sat_radar_combined.png
+            :width: 500 px
+
+To do this, users should first run :doc:`feature_detection_overview` with a dataset that contains latitude and longitude coordinates, such that they appear in the output dataframe from Feature Detection. Next, use :func:`tobac.utils.transform_feature_points` to transform the feature dataframe into the new coordinate system. Finally, use the output from :func:`tobac.utils.transform_feature_points` to run segmentation with the new data.
```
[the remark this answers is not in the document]
freemansw1: This actually wasn't properly set up for 3D, but I have now resolved that and have included that in the documentation here.

tobac/tests/test_internal_utils.py
```diff
@@ -0,0 +1,33 @@
+import tobac.utils.internal as internal_utils
```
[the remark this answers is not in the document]
freemansw1: Good catch. I've consolidated these.

tobac/utils/general.py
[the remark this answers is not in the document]
freemansw1: I have hopefully addressed this the best we can in the docs; let me know if more clarification is needed.

tobac/utils/general.py
```diff
@@ -553,3 +554,129 @@ def combine_tobac_feats(list_of_feats, preserve_old_feat_nums=None):
...
+        )
+        * 1000.0
+    )
+    ret_features = ret_features.where(dist_apart < max_space_away, drop=True)
+
+    # force times to match, where appropriate.
+    if "time" in new_dataset.coords:
+        # this is necessary due to the iris/xarray/pandas weirdness that we have.
+        old_feat_times = ret_features["time"].astype("datetime64[s]")
+        closest_times = np.min(np.abs(old_feat_times - new_dataset["time"]), axis=1)
```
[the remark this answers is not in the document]
freemansw1: I've switched to your suggestion, thanks!

tobac/utils/general.py
freemansw1: Yes, I agree that there are issues on either side. I am inclined to leave it for now, though? It's entirely possible I'll have to optimize this for some upcoming work, though.

freemansw1:
@JuliaKukulies @w-k-jones this is ready for re-review. @lettlini how can I/we add an additional few dependencies for the notebooks? The tutorial I include here requires PyArt and a few others.

doc/transform_segmentation.rst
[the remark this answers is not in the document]
JuliaKukulies: Thanks for fixing this - the page looks great!  However, the API reference does still not show up correctly. Any ideas why?

tobac/utils/internal.py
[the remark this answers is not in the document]
JuliaKukulies: Ah OK, I see now that there are other functions in the process that work fine without lat/lon, so let it be like this! Maybe through a warning or is this not really necessary?

doc/transform_segmentation.rst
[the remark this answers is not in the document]
JuliaKukulies: Perfect, thanks!

tobac/utils/general.py
JuliaKukulies: All right, that is OK for me.

tobac/utils/general.py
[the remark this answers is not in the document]
JuliaKukulies: Looks good in the documentation. And as discussed, we will address this issue in 1.6!

tobac/utils/general.py
```diff
@@ -553,3 +554,129 @@ def combine_tobac_feats(list_of_feats, preserve_old_feat_nums=None):
...
     return combined_df
+
+
+@internal_utils.irispandas_to_xarray
+def transform_feature_points(
+    features,
+    new_dataset,
+    latitude_name="auto",
+    longitude_name="auto",
+    max_time_away=datetime.timedelta(minutes=1),
```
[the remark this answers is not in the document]
JuliaKukulies: OK, setting the default to 0 also makes sense actually.

lettlini:
@freemansw1 I think it would be a good idea to add a separate step for installing packages that are not tobac dependencies. We should move `pytables` there as well.

To do this you would have to add a `step` to the action files, e.g.
```
- name: Install non-tobac dependencies
       run: |
           conda install -c conda-forge --yes pytables arm_pyart
```

Alternatively, you could just add the packages required by the notebook to the existing `Install tobac dependencies` step.

Let me know if you need any further help!

w-k-jones:
@freemansw1 @lettlini I don't think including pyart as a dependency is a good idea. When I try installing a clean environment with the tobac requirements + pyart it falls back to python `3.9`, so  it appears there are already some conflicts.

Would it be possible to create a file of the gridded radar output from pyart, upload this to zenodo and have the example notebook download and use this in the same manner as the original examples? We could also include the pyart code used to create the gridded dataset in a comment/markdown text to demonstrate how it was done

freemansw1:
> @freemansw1 @lettlini I don't think including pyart as a dependency is a good idea. When I try installing a clean environment with the tobac requirements + pyart it falls back to python `3.9`, so it appears there are already some conflicts.
> 
> Would it be possible to create a file of the gridded radar output from pyart, upload this to zenodo and have the example notebook download and use this in the same manner as the original examples? We could also include the pyart code used to create the gridded dataset in a comment/markdown text to demonstrate how it was done

I am happy to switch to that, but I think the downside there is that the user can't just download and run the code (which I see as a pretty big downside). @lettlini is there a way to make it notebook-specific, as a compromise? 

To be clear- just talking about adding pyart as a dependency for notebooks, not for the project.

tobac/tests/test_utils.py
w-k-jones, line 1: **Suggestion:** Test cases for features removed by `max_space_away` and `max_time_away` parameters

tobac/utils/general.py
w-k-jones: Would this function be called more than once for a single dataset? Would it be possible instead to do the accounting at the end and raise a single warning message informing the user that some features have been dropped?

tobac/utils/internal.py
```diff
@@ -454,16 +455,21 @@ def find_vertical_axis_from_coord(variable_cube, vertical_coord="auto"):
...
+        "altitude",
+        "geopotential_height",
+    ]
+
+    if isinstance(variable_cube, iris.cube.Cube):
+        list_coord_names = [coord.name() for coord in variable_cube.coords()]
+    elif isinstance(variable_cube, xr.Dataset) or isinstance(
+        variable_cube, xr.DataArray
+    ):
+        list_coord_names = variable_cube.coords
```
w-k-jones, line 470: **Suggestion:** Should we wrap this function with `@irispandas_to_xarray` to avoid checking internally?

tobac/utils/internal.py
```diff
@@ -674,3 +680,46 @@ def find_hdim_axes_3D_iris(field_in, vertical_coord=None, vertical_axis=None):
         all_axes[np.logical_not(np.isin(all_axes, [time_axis, vertical_coord_axis]))]
     )
     return output_vals
+
+
+@irispandas_to_xarray
+def detect_latlon_coord_name(in_dataset, latitude_name="auto", longitude_name="auto"):
```
w-k-jones, line 686: **Suggestion:** Would `None` be a better choice than `auto` for the default value and to automatically detect the coord name?

tobac/utils/internal.py
```diff
@@ -674,3 +680,46 @@ def find_hdim_axes_3D_iris(field_in, vertical_coord=None, vertical_axis=None):
...
+                out_lat = test_lat_name
+                break
+    if longitude_name.lower() != "auto":
+        if longitude_name in in_dataset.coords:
+            out_lon = longitude_name
+    else:
+        for test_lon_name in test_lon_names:
+            if test_lon_name in in_dataset.coords:
+                out_lon = test_lon_name
+                break
```
w-k-jones, line 724: **Question:** Do we want this to be case sensitive? It could be modified to detect coord names regardless of case (e.g. `"Lon"`, `"LON"`) but might be a bit more effort

w-k-jones (commented):
Looking good. I have still have a few concerns regarding the user understanding what's going on beneath the hood. This can be left to a later refactor if it's too much work to change quickly.

**Question:** What happens in segmentation if two features are assigned to the same point in the new dataset?

w-k-jones:
> 
> @w-k-jones may I ask in what order you installed Python and the other dependencies? Could you try the following:
> 
>     1. create a new virtual environment
> 
>     2. install the latest python version
> 
>     3. install all other packages in a separate step

Tested on two systems (OSX and linux), in both cases creating a new environment including `requirements.txt` and `arm_pyart` forces `python=3.9`, creating with `requirements.txt` and `python=3.11` and then trying to install `arm_pyart` fails to solve the environment. Solving with `mamba` works for both cases and installs with `python=3.11.3`

w-k-jones:
Even more complexity, on it's own `conda create -n test_pyart -c conda-forge --yes --dry-run arm_pyart` solves with `python=3.10.9`. It's a clash with `iris` that causes it to solve to 3.9...

tobac/utils/general.py
freemansw1: This is a good point, and not an idea that I had considered. I've added a warning that runs at the end. I have made it optional (but on by default).

doc/transform_segmentation.rst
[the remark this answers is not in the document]
freemansw1: This has been addressed in #297 , so will resolve this and let the conversation continue there.

tobac/utils/internal.py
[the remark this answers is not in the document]
freemansw1: I think this isn't necessary, as this should only be used internally, and the various functions that use it can decide to warn/not warn as appropriate.

tobac/utils/general.py
```diff
@@ -551,3 +556,146 @@ def combine_tobac_feats(list_of_feats, preserve_old_feat_nums=None):
...
+@internal_utils.irispandas_to_xarray
+def transform_feature_points(
+    features,
+    new_dataset,
+    latitude_name="auto",
+    longitude_name="auto",
+    altitude_name="auto",
+    max_time_away=datetime.timedelta(minutes=0),
+    max_space_away=0,
+    max_vspace_away=0,
```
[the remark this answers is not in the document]
freemansw1: This is a good point, and we typically try to avoid surprising user behavior. I've switched these to `None` and have added the appropriate logic to allow for no time/space maximum to be used.

tobac/utils/general.py
```diff
@@ -551,3 +556,146 @@ def combine_tobac_feats(list_of_feats, preserve_old_feat_nums=None):
...
+    if len(lat_vals_new.shape) != len(lon_vals_new.shape):
+        raise ValueError(
+            "Cannot work with lat/lon coordinates of unequal dimensionality"
+        )
+
+    # the lat/lons must be a 2D grid, so if they aren't, make them one.
+    if len(lat_vals_new.shape) == 1:
+        lon_vals_new, lat_vals_new = np.meshgrid(lon_vals_new, lat_vals_new)
+
+    # we have to convert to radians because scikit-learn's haversine
```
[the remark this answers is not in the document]
freemansw1: Generally agreed long-term, but we have to be somewhat careful as the `sklearn` haversine functions are so fast that they may often be preferrable.

tobac/utils/general.py
```diff
@@ -551,3 +556,146 @@ def combine_tobac_feats(list_of_feats, preserve_old_feat_nums=None):
...
+    # requires that the input be in radians.
+    flat_lats = np.deg2rad(lat_vals_new.flatten())
+    flat_lons = np.deg2rad(lon_vals_new.flatten())
```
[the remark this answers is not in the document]
freemansw1: Good call!

tobac/tests/test_utils.py
freemansw1: I believe that this is addressed in `test_transform_feature_points` already, but perhaps I am misunderstanding your suggestion here.

tobac/utils/internal.py
freemansw1: Doing this caused issues with some of the other functions, so I think debugging/updating this is outside the scope of this PR.

tobac/utils/internal.py
freemansw1: I think this is generally right. Changing this to be consistent throughout has been/will be a headache. I think we should do this correctly though, so will work on this.

tobac/utils/internal.py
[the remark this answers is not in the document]
JuliaKukulies: OK, agreed.

tobac/utils/internal.py
freemansw1: Ok, I switched "auto" to `None` throughout for all uses.

tobac/utils/internal.py
freemansw1: I am disinclined from making this case insensitive as it could lead to inconsistent results if a dataset has both `Lon` and `LON` (which seems unlikely, but definitely possible in some datasets).

freemansw1:
OK, I think I've resolved all of the review comments again. Ready for re-review @JuliaKukulies @w-k-jones .

JuliaKukulies (approved):
The changes look good to me @freemansw1, thanks for your hard work on this! 

The only thing that is not entirely clear for me yet is our dependency solution. So we use `example_requirements.txt` for the notebooks now? Somehow this seems to cause a fail in the notebook check workflow, but related to s3fs?

freemansw1:
> The only thing that is not entirely clear for me yet is our dependency solution. So we use `example_requirements.txt` for the notebooks now? Somehow this seems to cause a fail in the notebook check workflow, but related to s3fs?

Yes, it appears that it may be an issue with the pyart (?) dependencies not installing s3fs and fsspec versions that are compatible with each other. I've resolved this, but an actual bug was found, so I will need to resolve that next.

w-k-jones (approved):
Everything looks good, thanks for making the changes! Happy for this to be merged

PR #285, opened 2023-05-24
Allow user to select marker for unsegmented points

### LINKED ISSUE #62: Cell IDs are floats, while feature IDs are ints

freemansw1:
Resolves #62 and allows the user to select the marker for unsegmented points. This actually accomplishes two goals, as it reverts the default for segmentation (that all unsegmented points are `0`) back to what it was before the introduction of PBCs into this release. 

### DIFF
--- a/tobac/segmentation.py
+++ b/tobac/segmentation.py
@@ -320,6 +320,8 @@ def segmentation_timestep(
     PBC_flag="none",
     seed_3D_flag="column",
     seed_3D_size=5,
+    below_threshold_num=0,
+    above_threshold_unseg_num=0,
 ):
     """Perform watershedding for an individual time step of the data. Works
     for both 2D and 3D data
@@ -377,7 +379,10 @@ def segmentation_timestep(
         seed area for each dimension separately. Note: we recommend the use
         of odd numbers for this. If you give an even number, your seed box will be
         biased and not centered around the feature.
-
+    below_threshold_num: int
+        the marker to use to indicate a segmentation point is below the threshold.
+    above_threshold_unseg_num: int
+        the marker to use to indicate a segmentation point is above the threshold but unsegmented.
 
     Returns
     -------
@@ -441,6 +446,9 @@ def segmentation_timestep(
         raise ValueError(
             "Segmentation routine only possible with 2 or 3 spatial dimensions"
         )
+
+    if below_threshold_num > 0 or above_threshold_unseg_num > 0:
+        raise ValueError("Below/above threshold markers must be <=0")
 
     # copy feature dataframe for output
     features_out = deepcopy(features_in)
@@ -986,6 +994,9 @@ def segmentation_timestep(
 
     # Finished PBC checks and new PBC updated segmentation now in segmentation_mask.
     # Write resulting mask into cube for output
+    segmentation_mask_cpy = copy.deepcopy(segmentation_mask)
+    segmentation_mask[segmentation_mask_cpy == 0] = above_threshold_unseg_num
+    segmentation_mask[segmentation_mask_cpy == -1] = below_threshold_num
     segmentation_out.data = segmentation_mask
 
     # count number of grid cells associated to each tracked cell and write that into DataFrame:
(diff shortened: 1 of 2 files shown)


tobac/segmentation.py
```diff
@@ -986,6 +994,9 @@ def segmentation_timestep(
 
     # Finished PBC checks and new PBC updated segmentation now in segmentation_mask.
     # Write resulting mask into cube for output
+    segmentation_mask_cpy = copy.deepcopy(segmentation_mask)
+    segmentation_mask[segmentation_mask_cpy == 0] = above_threshold_unseg_num
+    segmentation_mask[segmentation_mask_cpy == -1] = below_threshold_num
```
w-k-jones, line 999: **Suggestion:** Rather than copying the entire array, we could instead save the masks first. This may save some memory as the boolean arrays should use less

```suggestion
    wh_below_threshold = segmentation_mask == -1
    wh_unsegmented = segmentation_mask == 0
    segmentation_mask[wh_unsegmented] = above_threshold_unseg_num
    segmentation_mask[wh_below_threshold] = below_threshold_num
```

tobac/segmentation.py
```diff
@@ -320,6 +320,8 @@ def segmentation_timestep(
     PBC_flag="none",
     seed_3D_flag="column",
     seed_3D_size=5,
+    below_threshold_num=0,
+    above_threshold_unseg_num=0,
```
w-k-jones, line 324: **Suggestion:** These parameter names are a bit unwieldy, and it may be difficult to change them after release. We could use a similar style to tracking which uses `cell_number_unassigned`, e.g. `segment_number_unassigned` and `segment_number_below_threshold`

w-k-jones (commented):
Thanks for making a PR to solve this issue! Neither of these comments are blocking, but would be nice to address before merging (particularly regarding parameter naming)

tobac/segmentation.py
freemansw1: Great suggestion, the names I came up with were bad.

tobac/segmentation.py
freemansw1: Good point. I've switched to bool arrays.

w-k-jones (approved):
Great! Thanks for making these changes

tobac/segmentation.py
```diff
@@ -1079,6 +1091,8 @@ def segmentation(
     PBC_flag="none",
     seed_3D_flag="column",
     seed_3D_size=5,
+    below_threshold_num=0,
```
JuliaKukulies, line 1094: you forgot to change the names here as well :)

tobac/segmentation.py
```diff
@@ -1202,6 +1220,8 @@ def segmentation(
             PBC_flag=PBC_flag,
             seed_3D_flag=seed_3D_flag,
             seed_3D_size=seed_3D_size,
+            segment_number_unassigned=above_threshold_unseg_num,
+            segment_number_below_threshold=below_threshold_num,
```
JuliaKukulies, line 1224: see above comment

JuliaKukulies (commented):
This looks great, @freemansw1! I just point out two locations where you seem to have forgotten the suggested name change by @w-k-jones. Once this is fixed, we are good to go!

tobac/segmentation.py
freemansw1: Good catches!

freemansw1:
Thanks! will merge after CI finishes.

Issue #153, opened 2022-07-02
Saving more information on data distribution within feature 


JuliaKukulies:
Currently, the center point locations (`lat, lon`/ `x,y`) the threshold (`threshold_value`) and the number of grid cells within that threshold (`num`) are the only info that we save about individual features (see [Feature detection output](https://tobac--150.org.readthedocs.build/en/150/feature_detection_output.html) ). 

It might be useful to add some bulk statistics about the datapoints within each feature/ segmented area (e.g. min, max, mean, percentiles, optionally the sum of all values, etc) in the feature and segmentation dataframes. This idea has been originally suggested in issue #18 and discussed with in #52, but I splitted issue #18, because this is a separate task. @freemansw1 suggested that this could be something for either v1.4.x or v1.5x and it may be good to wait until we have moved completely to xarray. 

Some things that need to be discussed before working on a PR: 
- In addition to bulk statistics, can we save the location of all feature points (as suggested by @freemansw1 in #52) ? Maybe optionally create a mask file just as the one we create for the segmentation? 

- Is it smarter to implement this as a postprocessing step or within the feature detection/ segmentation code? It would certainly be more computationally efficient and fairly easy to save this information directly within the feature detection/segmentation. However, it could be an advantage to make it a postprocessing step, because this would allow users to derive the same information using a different dataset (e.g. deriving precipitation statistics within cloud mask from a brightness temperature based tracking). 

- Maybe it would also be useful to have a function that updates the segmentation mask based on a modified feature dataframe. That way, one could easily use these additional statistics to filter out cells with specific properties and make sure that the mask files only contain those. I can see a lot of use cases for this, but could also be a postprocessing task that users should take care of themselves.

freemansw1:
Thanks for splitting off this issue! Here are some of my thoughts:

> In addition to bulk statistics, can we save the location of all feature points (as suggested by @freemansw1 in https://github.com/tobac-project/tobac/issues/52) ? Maybe optionally create a mask file just as the one we create for the segmentation?
I am inclined to wait for v2.x for this, as doing this with Iris sounds like a headache. We could push to get this in v1.x, but I think the best way to implement this on the backend would be to run it through xarray and then only convert to Iris at the end to output. I'm not sure how many people would use this feature, too, and we would need to document it thoroughly to avoid confusion with segmentation. 

[...]

PR #293, opened 2023-06-05
Let users optionally derive bulk statistics of the data points belonging to each feature 

JuliaKukulies:
Finally a PR that solves  #153 by implementing an optional extraction of statistics on the data points belonging to each feature. This is done in the segmentation step and produces additional columns in the segmentation output dataframe, if called by the user.

Added tests and updated example notebook and documentation accordingly. 

As discussed in #153, a follow-up on this after the transition to `xarray` will be to save not only the bulk statistics of features but to save the locations of all grid feature points. This is indirectly done in the mask files that can be combined with the actual data, but saving all feature data points would facilitate this analysis step for the user. 

* [x] If you have introduced a new functionality, have you added adequate unit tests? `test_get_statistics() `and `test_segmentation_multiple_features()`
* [x] If you have introduced a new functionality, have you added an example notebook? modified existing notebook
* [x] Have the newest changes from this branch been merged? merged in latest changes from `RC_v1.5.0`


w-k-jones:
Cool feature! Had a quick look through and it mostly looks good. Couple of questions/suggestions before I do a full review:

1. Can this be run offline? If we implement a function to call this offline (i.e. calculate stats for all features/segments that are previously detected) it might make for a nicer implementation in `segmentation_timestep`. I can see why this is also nice to include within the `segmentation` process as it removes the need for multiple loading for disk for large datasets.
2. Performance. How long does this take to run? Having a look, it loops over each feature and does a comparison to find each region individually, which can be quite slow. I have some faster alternatives for larger datasets that I can include if you're happy
3. Bulk statistics on feature detection. Can we add it in the same way?
4. Can we implement this in a way in which the user can supply a custom function to calculate over each segmented feature, rather than the current default statistics?

Most of these are beyond the scope of this PR, so more interesting thoughts raised by it than requirements!

Also `segmentation_timestep` is now 750 lines long 😬

JuliaKukulies:
Thanks for the great feedback, @w-k-jones! I really like your suggestions and I think that most of them could actually be implemented as part of this PR. Given that this is for `v1.6` we are not in a hurry, so I can work more on this with your input! 

>     1. Can this be run offline? If we implement a function to call this offline (i.e. calculate stats for all features/segments that are previously detected) it might make for a nicer implementation in `segmentation_timestep`. I can see why this is also nice to include within the `segmentation` process as it removes the need for multiple loading for disk for large datasets.

Not entirely sure I understand what you mean here. Do you mean if it is possible to run as a postprocessing step rather than during the segmentation?  

>     2. Performance. How long does this take to run? Having a look, it loops over each feature and does a comparison to find each region individually, which can be quite slow. I have some faster alternatives for larger datasets that I can include if you're happy

Note that we do the looping over features anyhow in the current code to get `ncells`. But you are right that it is not so effective to extract the right feature from the segmentation mask in each iteration. Yes, please provide me with some faster alternatives for this :) 

>     3. Bulk statistics on feature detection. Can we add it in the same way?

Sure thing! Are you thinking the bulk statistics for the feature with the most extreme threshold (since this is the one ending up in the output dataframe)? Or would it make more sense to get the statistics for the feature that was first detected (the least extreme) or all features within features? 

>     4. Can we implement this in a way in which the user can supply a custom function to calculate over each segmented feature, rather than the current default statistics?

Yes, good point, that makes sense! But we should allow for an arbitrary number of custom functions so that multiple statistics can be derived from one segmentation process, right? 

> Most of these are beyond the scope of this PR, so more interesting thoughts raised by it than requirements!

> Also `segmentation_timestep` is now 750 lines long grimacing

Haha, yes definitely not ideal :) Hope that we can shorten this down or implement elsewhere following your points 1-4.

lettlini (changes requested):
Thanks @JuliaKukulies for adding this!

I have only found a few minor things that should be looked at again. All in all very good work!

JuliaKukulies:
> Thanks @JuliaKukulies for adding this!
> 
> I have only found a few minor things that should be looked at again. All in all very good work!

Thanks for the quick and clear review, @lettlini ! Really good points - I addressed them in my latest commit.

I might ask you for a re-review when I have considered some of the points that concern the main design made by @w-k-jones . I think some at least make sense to address in this PR

w-k-jones:
> > ```
> > 1. Can this be run offline? If we implement a function to call this offline (i.e. calculate stats for all features/segments that are previously detected) it might make for a nicer implementation in `segmentation_timestep`. I can see why this is also nice to include within the `segmentation` process as it removes the need for multiple loading for disk for large datasets.
> > ```
> 
> Not entirely sure I understand what you mean here. Do you mean if it is possible to run as a postprocessing step rather than during the segmentation?
> 
Yes, sorry for being unclear! Running as a postprocessing step is indeed what I meant.

> > ```
> > 2. Performance. How long does this take to run? Having a look, it loops over each feature and does a comparison to find each region individually, which can be quite slow. I have some faster alternatives for larger datasets that I can include if you're happy
> > ```
> 
> Note that we do the looping over features anyhow in the current code to get `ncells`. But you are right that it is not so effective to extract each feature from the segmentation mask per iteration. Yes, please provide me with some faster alternatives for this :)
> 
Here's the function I use for calculating statistics over labelled functions:
https://github.com/w-k-jones/tobac-flow/blob/45b7541f0358fac670b94b78603935a15127d762/src/tobac_flow/utils/label_utils.py#L56-L100
It's similar to `scipy.ndimage.labeled_comprehension` but a bit more adaptable, including the ability to pass multiple fields to use as inputs to the functions for e.g. calculating weighted averages. You can also give it a function with multiple outputs, so we could calculate all statistics at once.
This approach is faster than doing a `groupby` over labels.
I've been calculating a lot a bulk properties of tracked objects in my own work, so can provide some examples if that would help! :)

> > ```
> > 3. Bulk statistics on feature detection. Can we add it in the same way?
> > ```
> 
> Sure thing! Are you thinking just the bulk statistics for the feature with the most extreme threshold (since this is the one ending up in the output dataframe)? Or would it make more sense to get the statistics for the feature that was first detected (the least extreme) or all features within features?
> 
I think just the most extreme threshold, as it would be difficult to do anything else without more complex logic/actually performing segmentation as well!

> > ```
> > 4. Can we implement this in a way in which the user can supply a custom function to calculate over each segmented feature, rather than the current default statistics?
> > ```
> 
> Yes, good point, that makes sense! But we should allow for an arbitrary number of custom functions so that multiple statistics can be derived from one segmentation process, right?
> 
Yes, we should allow for functions that return multiple values (so e.g. calculate mean and standard deviation), but also a list of different functions. This could be something added at a later date however

> > Most of these are beyond the scope of this PR, so more interesting thoughts raised by it than requirements!
> 
> > Also `segmentation_timestep` is now 750 lines long grimacing
>
> Haha, yes definitely not ideal :) Hope that we can shorten this down or implement elsewhere following your points 1-4.
> 
To be fair this is mainly due to the PBC and 3D changes, I just noticed it when looking at the code now! I'm sure with the switch to `xarray` in v1.6 we can do some refactoring and make it all a bit more manageable!

JuliaKukulies:
Great, thanks @w-k-jones for more helpful input and clarifications! Yes, if you could point me to some of your examples, that would be great! Your function for calculating statistics over labeled regions helps already a lot.

JuliaKukulies:
I have been working on this some more and am ready to get reviews again! Apologies in advance that this PR is larger than expected because I also worked on the tests and example notebooks for this new feature. 

I have changed the function for the bulk statistics quite a lot based on the example provided by @w-k-jones. As discussed in the dev meeting, users can now provide an arbitrary number of statistic functions and their respective column names for the output feature dataframe. It is also possible to provide input parameters for these functions since for some of them it is required (e.g. `np.percentile`). 

I have implemented the optional calculation of bulk statistics in both the feature detection and segmentation. In addition, it is possible to run it offline, i.e. as a postprocessing step given a segmentation mask. 

There might still be things we wanna change and I appreciate your input, especially on the following points: 
- Is the input dictionary format for `tobac.utils.general.get_statistics()` too complicated (given that you can provide tuple with function and input parameters instead of just a function) and if yes, what would be a more straight forward way? 
- Can we avoid looping through the rows in the feature output dataframe when assigning the output of the statistic functions ? I have not found a good solution for this given that the output for some statistics might be array-like rather than a scalar and given that we might not want to require that the all feature labels of the output dataframe are found in the respective mask. 
- Should we keep the notebooks with the real data or should we rather use simplified cases for the RTD page?

w-k-jones:
Sorry for the delay, just had a chance to look through and it looks really good! I'll try it out on some examples next week and finish my review

JuliaKukulies:
> Sorry for the delay, just had a chance to look through and it looks really good! I'll try it out on some examples next week and finish my review

I know it is quite a lot to look at, so thank you for taking the time for this @w-k-jones! Looking forward to your review!

w-k-jones:
Ok, I've had a chance to test this out with the MCSMIP work and have found a few situations in which I'm getting the wrong answer. I have some ideas of what might be causing this, are you happy if I push any fixes directly to your branch @JuliaKukulies ?

JuliaKukulies:
> Ok, I've had a chance to test this out with the MCSMIP work and have found a few situations in which I'm getting the wrong answer. I have some ideas of what might be causing this, are you happy if I push any fixes directly to your branch @JuliaKukulies ?

Oh yes, please feel free to push any fixes directly to my branch! Do you mean you the code is working but you get a wrong number for a statistic when you double-check a particular feature? If that is the case and you know in what cases this happens, we might want to also change the tests accordingly.

w-k-jones:
> > Ok, I've had a chance to test this out with the MCSMIP work and have found a few situations in which I'm getting the wrong answer. I have some ideas of what might be causing this, are you happy if I push any fixes directly to your branch @JuliaKukulies ?
> 
> Oh yes, please feel free to push any fixes directly to my branch! Do you mean the code is working but you get a wrong number for a statistic when you double-check a particular feature? If that is the case and you know in what cases this happens, we might want to also change the tests accordingly.

Yes, I'm getting the wrong answer. I think it might be to do with a problem in the indexing, either when selecting the labelled region or assigning the results to the DataFrame, which I don't think would come up in the tests as they use contiguous region labels. I'll update as I dig into it a bit more

JuliaKukulies:
> > > Ok, I've had a chance to test this out with the MCSMIP work and have found a few situations in which I'm getting the wrong answer. I have some ideas of what might be causing this, are you happy if I push any fixes directly to your branch @JuliaKukulies ?
> > 
> > 
> > Oh yes, please feel free to push any fixes directly to my branch! Do you mean the code is working but you get a wrong number for a statistic when you double-check a particular feature? If that is the case and you know in what cases this happens, we might want to also change the tests accordingly.
> 
> Yes, I'm getting the wrong answer. I think it might be to do with a problem in the indexing, either when selecting the labelled region or assigning the results to the DataFrame, which I don't think would come up in the tests as they use contiguous region labels. I'll update as I dig into it a bit more

All right great, yes sounds like it might be an issue with indexing and assigning the answers to the right location in the feature dataframe. I can also double-check again.

tobac/utils/general.py
```diff
@@ -631,6 +631,224 @@ def combine_feature_dataframes(
...
+                    ]
+                )
+
+            # add results of computed statistics to feature dataframe with column name given per func_dict
+            for idx, label in enumerate(np.unique(labels[labels > 0])):
+
+                # test if values are scalars
+                if not hasattr(stats[idx], "__len__"):
+                    # if yes, we can just assign the value to the new column and row of the respective feature
+                    features.loc[features[id_column] == label, stats_name] = stats[idx]
```
w-k-jones, line 758: When assigning the results to the dataframe, we loop over only values in the segment labels, but the stats array has elements for every contiguous integer between 1 and the highest label. If the segmentation is not contiguous (which is often the case) we end up assigning the wrong values to the results.

We can either fix this by assigning based on the label value, or limiting index so that we only calculate stats for labels that exist in the df. I'll look at implementing these and see if it fixes things

JuliaKukulies: Good catch, I totally missed that! I think it would be more elegant to define the index so that we only calculate statistics for labels that exist in the dataframe but I am fine with whatever is easier to implement. Thank you for your help with this!

w-k-jones: Have just pushed a quick fix for this

tobac/tests/test_segmentation.py
w-k-jones, line 1: We should add a test to check that the values output by statistics are correct

tobac/tests/test_utils.py
w-k-jones, line 1: Would it be possible to split the 2D and 3D into separate tests?

tobac/utils/general.py
```diff
@@ -631,6 +631,228 @@ def combine_feature_dataframes(
...
+    for tt in pd.to_datetime(segmentation_mask.time):
+        # select specific timestep
+        segmentation_mask_t = segmentation_mask.sel(time=tt).data
+        field_t = field.sel(time=tt).data
+
+        # make sure that the labels in the segmentation mask exist in feature dataframe
+        if (
+            np.intersect1d(np.unique(segmentation_mask_t), features.feature).size
+            > np.unique(segmentation_mask_t).size
+        ):
```
w-k-jones, line 837: I think this should be less than, rather than greater than
```suggestion
        if (
            np.intersect1d(segmentation_mask_t, features[id_column]).size
            < np.unique(segmentation_mask_t).size
        ):
```

However, segmentation does not necessarily create a label for every feature (e.g. if the seed location is below the segmentation threshold value), so we should be careful about raising an exception here as it may break segmentation

tobac/utils/general.py
```diff
@@ -631,6 +631,228 @@ def combine_feature_dataframes(
...
+    if not features.feature.is_unique:
+        raise logging.warning(
+            "Feature labels are not unique which may cause unexpected results for the computation of bulk statistics."
+        )
+
+    # get bulk statistics for each timestep
+    for tt in pd.to_datetime(segmentation_mask.time):
+        # select specific timestep
+        segmentation_mask_t = segmentation_mask.sel(time=tt).data
+        field_t = field.sel(time=tt).data
```
w-k-jones, line 831: This only applies to the final field from the loop at line 817
```suggestion
        field_t = [field.sel(time=tt).data for field in fields]
```

We could do this in the call to `get_statistics` using parameter expansion

tobac/utils/general.py
```diff
@@ -631,6 +631,228 @@ def combine_feature_dataframes(
...
+            # get the statistics only for specified feature objects
+            if np.max(index) > np.max(labels):
+                raise ValueError("Index contains values that are not in labels!")
+
+        # Find which labels exist in features for output:
+        index_in_features = np.isin(index, features[id_column])
+
+        # set negative markers to 0 as they are unsegmented
+        labels[labels < 0] = 0
+        bins = np.cumsum(np.bincount(labels.ravel()))
```
w-k-jones, line 696: To avoid changing the labels array:
```suggestion
        bins = np.cumsum(np.bincount(np.maximum(labels.ravel(), 0)))
```

tobac/utils/general.py
```diff
@@ -631,6 +631,228 @@ def combine_feature_dataframes(
...
+                            *[
+                                field.ravel()[argsorted[bins[i - 1] : bins[i]]]
+                                for field in fields
+                            ]
+                        )
+                        if bins[i] > bins[i - 1]
+                        else default
+                        for i in index
+                    ]
+                )
```
w-k-jones, line 752: We can avoid duplicating this code by using functools.partial to wrap the keywords with the function call.

e.g.: 
```
if type(func_dict[stats_name]) is tuple:
    func = partial(func_dict[stats_name][0], **func_dict[stats_name][1])
else:
    func = func_dict[stats_name][0]
```

tobac/utils/general.py
```diff
@@ -631,6 +631,228 @@ def combine_feature_dataframes(
...
+    for field in fields:
+        if labels.shape != field.shape:
+            raise ValueError("Input labels and field do not have the same shape")
+
+    # mask must contain positive values to calculate statistics
+    if labels[labels > 0].size > 0:
+        if index is None:
+            index = range(
+                int(np.nanmin(labels[labels > 0])), int(np.nanmax(labels) + 1)
+            )
```
w-k-jones, line 685: We change the default values for index to the intersect of features and labels to reduce the amount of work performed

tobac/utils/general.py
```diff
@@ -631,6 +631,228 @@ def combine_feature_dataframes(
...
+            raise ValueError("Input labels and field do not have the same shape")
+
+    # warning when feature labels are not unique in dataframe
+    if not features.feature.is_unique:
+        raise logging.warning(
+            "Feature labels are not unique which may cause unexpected results for the computation of bulk statistics."
+        )
+
+    # get bulk statistics for each timestep
+    for tt in pd.to_datetime(segmentation_mask.time):
```
w-k-jones, line 828: Looping over the time dimension makes this a bit more complicated, but could be a good way to add chunking for dask support later

tobac/utils/general.py
w-k-jones, line 1: Would it be a good idea to split the bulk statistics functions into a separate file to avoid utils.general from getting too large?

w-k-jones (changes requested):
A really nice addition to tobac, and very useful for the MCSMIP project! I'm happy to make the changes I've suggested myself, but would appreciate your comments.

Also, we should probably merge the changes in the RC_v1.5.x branch as I think these are needed to solve the merge conflict without breaking things

tobac/utils/general.py
```diff
@@ -631,6 +631,228 @@ def combine_feature_dataframes(
     return combined_sorted
 
 
+def get_statistics(
+    labels: np.ndarray[int],
+    *fields: tuple[np.ndarray],
+    features: pd.DataFrame,
+    func_dict: dict[str, tuple[Callable]] = {"ncells": np.count_nonzero},
```
w-k-jones, line 638: We could call this parameter `statistics` as that would follow the naming convention from scipy.stats.binned_statistic
```suggestion
    statistic: dict[str, Callable | tuple[Callable, dict]] = {"ncells": np.count_nonzero},
```

JuliaKukulies:
> A really nice addition to tobac, and very useful for the MCSMIP project! I'm happy to make the changes I've suggested myself, but would appreciate your comments.
> 
> Also, we should probably merge the changes in the RC_v1.5.x branch as I think these are needed to solve the merge conflict without breaking things

Many thanks for the useful comments and suggestions @w-k-jones ! I will work on these and merge in RC_v.1.5.x as you say

tobac/utils/general.py
JuliaKukulies: Happy with this fix, thank you!

tobac/utils/general.py
JuliaKukulies: I am down to do this! Any thoughts @freemansw1 since you worked with the restructuring of the `utils` module?

doc/bulk_statistics/notebooks/statistics_segmentation_example.ipynb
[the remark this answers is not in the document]
JuliaKukulies: Done

tobac/utils/general.py
JuliaKukulies: I agree that this is not optimal. Wait, do we even have to do that? I cant remember if I did that for a reason or not. Looking at your original code for applying functions over regions, it should actually work for a 3D/4D array, or am I missing something?

tobac/utils/general.py
JuliaKukulies: Nice idea, I have done that!

w-k-jones:
From discussion regarding type hints in #337, #341 we should make sure to add `from __future__ import annotations` and possibly consider using `from typing import Union` rather than `|` for union type hints to ensure back-compatibility with python versions before 3.10

freemansw1 (commented):
I haven't run this fully yet on all the datasets I want to, but I am very happy with this so far! Nice addition!

tobac/utils/general.py
```diff
@@ -631,6 +632,209 @@ def combine_feature_dataframes(
...
+                [
+                    func(
+                        *[
+                            field.ravel()[argsorted[bins[i - 1] : bins[i]]]
+                            for field in fields
+                        ],
+                    )
+                    if bins[i] > bins[i - 1]
+                    else default
+                    for i in index
```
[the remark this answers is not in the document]
JuliaKukulies: OK, great. I agree! Have created a submodule now

tobac/utils/general.py
JuliaKukulies: OK, great. I agree! Have created a submodule now

doc/index.rst
```diff
@@ -60,6 +60,14 @@ The project is currently being extended by several contributors to include addit
 
    merge_split
 
+
+.. toctree::
+   :caption: Compute bulk statistics
```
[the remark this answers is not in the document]
JuliaKukulies: I changed the header to `Compute bulk statistics online or in postprocessing` and then the subsections/notebooks are called `compute_statistics_during_feature_detection`, `compute_statistics_during_feature_detection` and `compute_statistics_postprocessing`. Do you think that is clear enough to indicate that both options offline and online computation of the statistics is possible? Happy to change the header again or to add a paragraph on the RTD page that clarifies these three options

JuliaKukulies:
> From discussion regarding type hints in #337, #341 we should make sure to add `from __future__ import annotations` and possibly consider using `from typing import Union` rather than `|` for union type hints to ensure back-compatibility with python versions before 3.10

I changed to `Union` for this PR

tobac/utils/general.py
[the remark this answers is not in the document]
JuliaKukulies: Can you clarify what you mean by subsetting in this case? :)  Are you referring to the indices?

JuliaKukulies:
@w-k-jones and @freemansw1 I think I addressed all your comments, so it would be nice if you could re-review whenever it is convenient for you! After we merge this PR, I will do ahead and prepare the 1.5.2 release with #337, #351 and #341 as we discussed in the last dev meeting

w-k-jones (approved):
Just a small change to replicate previous behaviour for the `ncells` when no segment is present, otherwise happy to merge

freemansw1 (approved):
Very nice change, @JuliaKukulies ! I've tested on a variety of datasets and seem to get reasonable results. I'm extremely happy that we're getting these changes in now.

JuliaKukulies:
> Very nice change, @JuliaKukulies ! I've tested on a variety of datasets and seem to get reasonable results. I'm extremely happy that we're getting these changes in now.

Thank you very much for checking this new feature with different datasets! I will go ahead and merge when the jupyter tests pass (which should be the case now that I have merged Wills last PR with the zenodo path changes)

tobac/utils/bulk_statistics.py
```diff
@@ -0,0 +1,215 @@
+"""
+Support functions to compute bulk statistics of features, either as a postprocessing step
+or within feature detection or segmentation. 
+
+"""
+
```
w-k-jones, line 6: Quick fix: missing `logging` import
```suggestion
import logging
```

w-k-jones:
Just found an annoying bug with missing segments in which all the features which don't have a label in the segment mask before the first present label will get the value `None` instead of the default. Will have a look into a quick fix...

JuliaKukulies:
> Just found an annoying bug with missing segments in which all the features which don't have a label in the segment mask before the first present label will get the value `None` instead of the default. Will have a look into a quick fix...

Thanks for these fixes! And ha, of course, I found another bug that let 
 the notebook check for `Example_Track_on_Radar_Segment_on_Satellite/track_on_radar_segment_on_satellite.ipynb` fail. The issue is was that `utils.transform_featurepoints` uses the `irispandas_to_xarray` decorator and `xr.DataArray.where() ` on the feature values which converted them from integers to floats. That was already the case before but now the segmentation step in the notebook fails because `utils.get_statistics()` needs the feature IDs to be integers

freemansw1:
> Do you want to have another quick look @freemansw1 @w-k-jones or are we ready to merge?
> 
I am happy for this to be merged. 

> And for the future: Would it make sense to implement tests that check the datatypes of the output data frames at least in all functions that modify the data frames?

Yes, we should probably introduce those as part of our typical suite.

PR #323, opened 2023-08-15
Merge `RC_v1.5_hotfix` into `main`

w-k-jones:
Hotfix v1.5.1 release for issue with strict thresholding found in #314 and #315, fixed by #316

* [x] Have you followed our guidelines in CONTRIBUTING.md? 
* [x] Have you self-reviewed your code and corrected any misspellings? 
* [x] Have you written documentation that is easy to understand?
* [x] Have you written descriptive commit messages? 
* [x] Have you added NumPy docstrings for newly added functions? 
* [x] Have you formatted your code using black? 
* [x] If you have introduced a new functionality, have you added adequate unit tests?
* [x] Have all tests passed in your local clone? 
* [x] If you have introduced a new functionality, have you added an example notebook?
* [x] Have you kept your pull request small and limited so that it is easy to review? 
* [x] Have the newest changes from this branch been merged?

PR #325, opened 2023-08-15
V1.5.1 release updates

w-k-jones:
Update version number, changelog and notebooks for v1.5.1 release

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 ### Tobac Changelog
 
+_**Version 1.5.1:**_
+
+**Bug fixes**
+
+- The `strict_thresholding` option in feature detection now works correctly for detecting minima, and produces the same results as without strict thresholding if the `n_min_threshold` is a scalar value [#316](https://github.com/tobac-project/tobac/pull/316)
+
+
 _**Version 1.5.0:**_
 
 **Enhancements for Users**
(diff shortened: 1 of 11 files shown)


JuliaKukulies (approved):
Looks good - thanks for taking care of this hotfix, @w-k-jones! I assume that we should merge this before #323. 

One thing I noticed is that you called the new hotfix branch `RC_v1.5_hotfix` which has the same branch pattern that we use for the release candidates (`RC_v*`). That is not a big deal at all, but I just wanted to point out this because it means that the branch falls under the branch protection rules  for release candidates (2 reviews) now.

freemansw1 (approved):
Thanks for doing this, Will.

w-k-jones:
> One thing I noticed is that you called the new hotfix branch `RC_v1.5_hotfix` which has the same branch pattern that we use for the release candidates (`RC_v*`). That is not a big deal at all, but I just wanted to point out this because it means that the branch falls under the branch protection rules for release candidates (2 reviews) now.

I did that specifically so that the branch protection rules would apply, but perhaps we should consider in future a different workflow for hotfixes?

JuliaKukulies:
> > One thing I noticed is that you called the new hotfix branch `RC_v1.5_hotfix` which has the same branch pattern that we use for the release candidates (`RC_v*`). That is not a big deal at all, but I just wanted to point out this because it means that the branch falls under the branch protection rules for release candidates (2 reviews) now.
> 
> I did that specifically so that the branch protection rules would apply, but perhaps we should consider in future a different workflow for hotfixes?

Ah OK :)  I thought that we had agreed at some point that one review suffices for hotfixes, but we can re-evaluate if this really should be the case.

PR #337, opened 2023-09-10
Add type hints to feature detection

### LINKED ISSUE #75: Introducing type hints
I have made really good experiences with using Python type hints in my recent python projects (similar size and number of contributors our current tobac...). 

It might be a good idea to have a look at it, to make it easier to develop and work with the code, especially when using higher level classes or if we think about packaging some things on our own tobac classes.

For those of you who are not familiar with type hints in Python:  
It basically extends type-agnostic Python  with typing familiar from many other programming languages (but not as strict.)

So an example of a function with type hints (just something stupid  I just made up...) would look like:
```
def some_function(input_data: pd.DataFrame, 
                              column: str,  
                              input_number: float) -> pd.Series:
     series_out=data_in['column']*input_number
     return series_out
```

As far as I understand, it is not enforced by Python when running the code, but helps a lot in development as most IDEs will point you to a wrong type or class while writing code using the function be in with regards to the input or with regards to the return values.
It is obiously in some way a duplication of the info in the docstrings, but in a more applicable way that is directly readable to an IDE

We could also start using it for new additions and then add it to the rest of the code (should be straight forward where there are already the right docstrings ;) )

Any thoughts or experiences?

Some more info:
https://www.python.org/dev/peps/pep-0484/
https://towardsdatascience.com/type-hints-in-python-everything-you-need-to-know-in-5-minutes-24e0bad06d0b

freemansw1:
Partially resolves #75 for feature detection only and adds type hints to all function calls. At the same time, I also went through and dealt with some of the easier issues that pylint found. 

I started this process in an effort to start to think through our xarray transition. It helped me think through *where* iris is actually used (which isn't often in feature detection). 

Note that to get `Literal` to work, we will have to increase our Python requirement to 3.8. I think this is reasonable given that 3.7 is EOL and is no longer available to be installed through `conda`. 3.7 first came out in 2018, but its most recent release was this June. There are two alternatives if we don't want to do this: 1. require the `typing_extensions` package. This is a new requirement, and all new requirements should be scrutinized, but given that it is from `python` itself, I think there is a low risk for inclusion. 2. make the types for our options just `str`, which I think is less preferred. 

In the absence of automated matrix testing, I've run tests on my local machine with Python 3.11 and 3.8, both of which pass. There are still a number of linting errors, but c'est la vie. 


freemansw1:
I decided to reverse course and add `typing_extensions` as a requirement to preserve 3.7 support and because I think it's highly likely that we will need the backporting of typing features at some point.

w-k-jones:
I'm pro removing python 3.7 support with the v1.6 update, but including `typing_extensions` is probably a good idea for now

freemansw1:
I've switched this to a draft PR until we get 1.5.1 out. `RC_v1.5.x` currently lags `RC_v1.5_hotfix`.

freemansw1:
With 1.5.1 out, I've marked this as ready for review. I think there's an argument to wait on this (and #341) until v1.6.0, but I think there is no harm in doing so with v1.5.2 as well.

tobac/feature_detection.py
```diff
@@ -16,35 +16,44 @@
    diverse datasets. Geoscientific Model Development,
    12(11), 4551-4570.
 """
-
+from __future__ import annotations
+from typing import Union
+import warnings
 import logging
+
+from typing_extensions import Literal
```
w-k-jones, line 24: I think we should consider moving the requirements to python 3.8 for the 1.6 release, which would remove this dependency

tobac/feature_detection.py
```diff
@@ -1350,20 +1393,20 @@ def feature_detection_multithreshold(
...
+    min_distance: float = None,
+    x_coordinate_name: str = None,
+    y_coordinate_name: str = None,
+    z_coordinate_name: str = None,
+    target: Literal["maximum", "minimum"] = "maximum",
+    PBC_flag: Literal["none", "hdim_1", "hdim_2", "both"] = "none",
+    min_h1: int = 0,
+    max_h1: int = 0,
+    min_h2: int = 0,
+    max_h2: int = 0,
```
w-k-jones, line 1408: are these variables `int` or can they be `float`? I've been using float values with PBCs for tracking, but I'm not sure if it's the same for feature detection

w-k-jones (commented):
Looks good, my only query is regarding the typing of the `min_h1`, `max_h1` etc... variables for PBCs

tobac/feature_detection.py
freemansw1: These variables cannot be `float`; they should be array indices.

w-k-jones (approved):

tobac/feature_detection.py
freemansw1: I guess the question is do we want to get this in 1.5.2 or wait for 1.6.0. I'm inclined to push it through and use the typing extensions, but I'm not committed to that plan.

w-k-jones: I'm happy to use `typing_extensions` for now and make a note to change it for 1.6

tobac/feature_detection.py
JuliaKukulies: Can we specify that these are array indices in the docstrings? I interpreted it as the actual dimension values/coordinates and I am surprised this did not change my results for the MCSMIP tracking now when I corrected it.

JuliaKukulies: Also, I realized that these are not required by the user when running the feature detection with PBC, while the tracking with PBC requires the user to define these (but that should be another issue), just as the docstrings.

tobac/feature_detection.py
JuliaKukulies: Agree with the above

JuliaKukulies (approved):

JuliaKukulies:
Looks very good - also nice that you addressed some pylint issues in the same go!

tobac/feature_detection.py
JuliaKukulies: Sorry for posting this here. I know this should be a separate issue, but I wanted to clarify first if I interpret it right.

harrietgilmour (approved):
I really like this addition! It all looks good and a great starting point for adding type hints to the rest of tobac (if that's still the plan).

freemansw1:
Updated now that we have merged in #293

freemansw1:
@JuliaKukulies did I get the type for the statistics right? If so, I will go ahead and merge.

JuliaKukulies:
> @JuliaKukulies did I get the type for the statistics right? If so, I will go ahead and merge.

Yes, looks good!

Issue #360, opened 2023-11-07
`time_cell_min` doesn't appear to work


freemansw1:
First reported by @smsaleeb. As of 1.5.1, it appears `time_cell_min` doesn't work. Working now to confirm the issue and on a PR.

freemansw1:
Never mind- issue on my end.

PR #361, opened 2023-11-07
Add tests for `time_cell_min` and update documentation for the parameter

freemansw1:
Follows from my mistake in #360 - this adds a specific test for time_cell_min and updates the documentation for the parameter in the tracking API docs to reflect the correct units (not minutes). 


JuliaKukulies (approved):
Approved, but a question to double-check:  Can `dt` really be provided in any different units than seconds? 

I am thinking of this line here where `dt` is related to `vmax`:
https://github.com/tobac-project/tobac/blob/09b5a733e0309d8dde47d7ceeccd273cdc8aa7dd/tobac/tracking.py#L234 

Maybe we should also add in the documentation then that `vmax` needs to be per same time unit as `dt`. Same is true for `dxy`

freemansw1:
> Approved, but a question to double-check: Can `dt` really be provided in any different units than seconds?
> 
> I am thinking of this line here where `dt` is related to `vmax`:
> 
> https://github.com/tobac-project/tobac/blob/09b5a733e0309d8dde47d7ceeccd273cdc8aa7dd/tobac/tracking.py#L234
> 
> Maybe we should also add in the documentation then that `vmax` needs to be per same time unit as `dt`. Same is true for `dxy`

This is a good point- these units all need to match (and communicating that without specifying units can be a headache). I wonder if we shouldn't just require that `dt` be in seconds, `dxy` in meters, and `vmax` in m/s? In theory these can all be in different units as long as they are consistent, but may be easier just to message the proper units.

JuliaKukulies:
> > Approved, but a question to double-check: Can `dt` really be provided in any different units than seconds?
> > I am thinking of this line here where `dt` is related to `vmax`:
> > https://github.com/tobac-project/tobac/blob/09b5a733e0309d8dde47d7ceeccd273cdc8aa7dd/tobac/tracking.py#L234
> > 
> > Maybe we should also add in the documentation then that `vmax` needs to be per same time unit as `dt`. Same is true for `dxy`
> 
> This is a good point- these units all need to match (and communicating that without specifying units can be a headache). I wonder if we shouldn't just require that `dt` be in seconds, `dxy` in meters, and `vmax` in m/s? In theory these can all be in different units as long as they are consistent, but may be easier just to message the proper units.

Yes, totally agree that it is a bit difficult to communicate this and that it therefore would be best to just stick to SI units

JuliaKukulies:
We discussed this with the whole developer group and agreed that it would be easiest to require SI units for all input parameters, i.e. seconds for times, meters for distances, m/s for speeds etc. In the future, it would maybe be smart to test for consistent units as well, but I think we will be OK for now with just updating the docstrings. Did you want a second reviewer for this as well @freemansw1 ?

freemansw1:
> We discussed this with the whole developer group and agreed that it would be easiest to require SI units for all input parameters, i.e. seconds for times, meters for distances, m/s for speeds etc. In the future, it would maybe be smart to test for consistent units as well, but I think we will be OK for now with just updating the docstrings. Did you want a second reviewer for this as well @freemansw1 ?

Yes, good call! Checking units can be a bit of a headache (we could go down the path of requiring units like metpy does, but I've found dealing with pint et al. to be cumbersome), but worth exploring for the future. For now, we can just document the required units.

tobac/tests/test_tracking.py
```diff
@@ -557,3 +559,94 @@ def test_untracked_nat():
...
+    cell_time_lengths: list[int]
+        Length that each cell appears for
+    min_time_length: int
+        time_cell_min value
+    expected_there: list[bool]
+        whether to expect the cell there.
+
+    """
+    # delta time automatically set by smallest difference between cell lengths
+    delta_time = int(np.min(np.diff(cell_time_lengths)))
```
snilsn, line 612: Why go through the struggle of this line and have to use floor division twice, when you could just set `num_frames` to the original cell time lengths (so that `delta_time` would be 1, no division needed) and then assert against the original cell time lengths again at the end?

snilsn (approved):
I'm convinced this does what it should, so I'm approving it. I only have one question/comment regarding readability, but I guess that doesn't really matter since it's just a test.

tobac/tests/test_tracking.py
freemansw1: Good point. I've updated the code just to set `delta_time` to 1.

freemansw1:
Ok, I've addressed @snilsn's comment and have gone through and specified units in the tracking documentation. @JuliaKukulies Can you check that before we merge?

JuliaKukulies:
> Ok, I've addressed @snilsn's comment and have gone through and specified units in the tracking documentation. @JuliaKukulies Can you check that before we merge?

Looks good, thanks for adding the units for the tracking documentation! 

Can we add the same for `dxy` and `dz` in the feature detection and segmentation? I could not directly push commits to this PR, otherwise I would have added them myself

freemansw1:
> > Ok, I've addressed @snilsn's comment and have gone through and specified units in the tracking documentation. @JuliaKukulies Can you check that before we merge?
> 
> Looks good, thanks for adding the units for the tracking documentation!
> 
> Can we add the same for `dxy` and `dz` in the feature detection and segmentation? I could not directly push commits to this PR, otherwise I would have added them myself

Good point. I've gone through and made updates to the documentation of both segmentation and feature detection.

PR #362, opened 2023-11-08
Update black integration to say what is wrong and fix python versions

freemansw1:
Coming from a frustrating experience developing #354, update our black integration to specify what it would change and also set the versions to ensure compatibility. We will need to update this file when we bump versions (on either end). 

### DIFF
--- a/.github/workflows/check_formatting.yml
+++ b/.github/workflows/check_formatting.yml
@@ -19,4 +19,4 @@ jobs:
         shell: bash -l {0}
         run:
           mamba install --quiet --yes --file requirements.txt black &&
-          black tobac --check
+          black tobac --check --diff -t py37 -t py38 -t py39 -t py310 -t py311 -t py312
--- a/.github/workflows/check_formatting.yml
+++ b/.github/workflows/check_formatting.yml
@@ -19,4 +19,4 @@ jobs:
         shell: bash -l {0}
         run:
           mamba install --quiet --yes --file requirements.txt black &&
-          black tobac --check --diff -t py37 -t py38 -t py39 -t py310 -t py311
+          black tobac --check --diff 
--- a/.github/workflows/check_formatting.yml
+++ b/.github/workflows/check_formatting.yml
@@ -19,4 +19,4 @@ jobs:
         shell: bash -l {0}
         run:
           mamba install --quiet --yes --file requirements.txt black &&
-          black tobac --check --diff -t py37 -t py38 -t py39 -t py310 -t py311 -t py312
+          black tobac --check --diff -t py37 -t py38 -t py39 -t py310 -t py311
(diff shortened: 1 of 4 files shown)


freemansw1:
Interesting that when we add specific versions, `black` comes up with some new things for us to fix. I'm happy to remove the specific versions (as long as we leave the `diff`). @JuliaKukulies You did the initial integration here; what do you think?

w-k-jones:
Interesting, I'm not sure why previous formatting jobs haven't caught these issues

freemansw1:
> Interesting, I'm not sure why previous formatting jobs haven't caught these issues

Yes when you run just `black tobac`, it doesn't have anything, but `black tobac -t py37 -t py38 -t py39 -t py310 -t py311` comes up with all sorts of stuff. Not a clue why. If we switched, we would need to update our `pre-commit` too I've just realized, and also everyone would need to update their IDEs (ugh).

JuliaKukulies:
> > Interesting, I'm not sure why previous formatting jobs haven't caught these issues
> 
> Yes when you run just `black tobac`, it doesn't have anything, but `black tobac -t py37 -t py38 -t py39 -t py310 -t py311` comes up with all sorts of stuff. Not a clue why. If we switched, we would need to update our `pre-commit` too I've just realized, and also everyone would need to update their IDEs (ugh).

Yeah, that seems cumbersome. I am having similar issues with another coding project where it seems like black does the formatting differently for the same version of black in combination with different python versions which lets me believe that the formatter is not so strict. But on the other hand, this causes the checks to fail so it is weird that it would be strict on the checking end of things while not being strict on the formatting itself. Let me have a more detailed look into this and find out if this is really the intended behavior.

freemansw1:
Okay, the specific issue it's cluing in on here is adding an extra comma after `**kwargs`, which was only supported in Python versions newer than 3.6 (https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#trailing-commas). In code where it detects Python >3.6, it has already added them with simply `black filename`, but if it doesn't detect Python 3.6, it will not reformat. Adding the version specifiers adds the trailing comma as a requirement.

.github/workflows/check_formatting.yml
```diff
@@ -19,4 +19,4 @@ jobs:
         shell: bash -l {0}
         run:
           mamba install --quiet --yes --file requirements.txt black &&
-          black tobac --check
+          black tobac --check --diff -t py37 -t py38 -t py39 -t py310 -t py311
```
JuliaKukulies, line 22:
```suggestion
          black tobac --check --diff 
```

JuliaKukulies (approved):

JuliaKukulies:
Looks goog to me, thanks for these changes, @freemansw1!

w-k-jones (approved):
Looks good, happy for this to be merged

PR #368, opened 2023-11-23
Update bulk statistics to use multiple fields correctly

w-k-jones:
Fixes `get_statistics_from_mask` to work with multiple field inputs correctly, resolving #364, and also solves the issue that was causing the jupyter notebook CI to fail

* [x] Have you followed our guidelines in CONTRIBUTING.md? 
* [x] Have you self-reviewed your code and corrected any misspellings? 
* [x] Have you written documentation that is easy to understand?
* [x] Have you written descriptive commit messages? 
* [x] Have you added NumPy docstrings for newly added functions? 
* [x] Have you formatted your code using black? 
* [x] If you have introduced a new functionality, have you added adequate unit tests?
* [x] Have all tests passed in your local clone? 
* [x] If you have introduced a new functionality, have you added an example notebook?
* [x] Have you kept your pull request small and limited so that it is easy to review? 
* [x] Have the newest changes from this branch been merged?

PR #372, opened 2023-11-29
Add PBC capability to `merge_split_MEST`

w-k-jones:
This turned into a general refactor of `merge_split_MEST`, with the following changes and additions:

1. Adds PBC capabilities, using the same `BallTree` distance search approach as used in feature detection.
2. Moves filtering by `frame_len` before calculating the minimum spanning tree, improving the number of merging/splitting cells linked particularly for long tracking time periods.
3. Changes the cell merging logic to use [scipy.sparse.csgraph.connected_components](https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.csgraph.connected_components.html#scipy.sparse.csgraph.connected_components), improving performance
4. Made sure all output arrays are int dtype, and started the `track` id from 1 rather than 0

I'm also going to look into adding a flag for whether a cell started with a split or ended with a merge, and which object it was merged into/split from


w-k-jones:
Keeping as draft until #368 is merged as this contains the same commits

kelcyno:
Oh this looks really interesting @w-k-jones I'll get started on it!

w-k-jones:
Ok, I ended up going down a bit of a rabbit hole there. I found that because the filtering for `frame_len` was carried out after the minimum euclidean spanning tree calculation, valid links were being removed because a _spatially_ closer cell at another point in time was being removed and therefore removing the edges between two neighbouring cells. This had more of an impact the longer the time period being tracked. When I swapped to filtering by `frame_len` before calculating the MEST it massively increased the number of linked cells, which then caused the rest of the merge/split routine to run slower. I've been experimenting with [scipy.sparse.csgraph.connected_components](https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.csgraph.connected_components.html#scipy.sparse.csgraph.connected_components) recently, and this seemed like a really good application for linking the neighbouring cells into connected tracks

freemansw1:
I haven't started reviewing this yet, but I wonder if we shouldn't try to push this to v1.5.3? I'm starting to get antsy about v1.5.2 coming out.

tobac/feature_detection.py
```diff
@@ -1461,7 +1461,7 @@ def filter_min_distance(
     target: {'maximum', 'minimum'}, optional
         Flag to determine if tracking is targetting minima or maxima in
         the data. Default is 'maximum'.
-    PBC_flag : str('none', 'hdim_1', 'hdim_2', 'both')
+    PBC_flag : str('none', 'hdim_1', 'hdim_2', 'both'), optional
```
freemansw1, line 1464: I'm not sure what style is required here, but I think we should move from `optional` to say that the default is `none`.

tobac/merge_split.py
```diff
@@ -22,6 +44,8 @@ def merge_split_MEST(TRACK, dxy, distance=None, frame_len=5):
         The x/y grid spacing of the data.
         Should be in meters.
 
+    dz : float, optional
```
freemansw1, line 47: In the rest of the 3D code, we support both constant vertical grid spacing and user-specified varying vertical grid spacing. We can either require constant vertical grid spacing for this code (less ideal) and check that the user didn't pass in an array-like, or update the code to support both (if it doesn't already). I'm happy enough with either outcome.

tobac/merge_split.py
```diff
@@ -32,6 +56,29 @@ def merge_split_MEST(TRACK, dxy, distance=None, frame_len=5):
         Threshold for the maximum number of frames that can separate the end of cell and the start of a related cell.
         Default is five (5) frames.
 
+    cell_number_unassigned: int, optional
+        Value given tp unassigned/non-tracked cells by tracking. Default is -1
+
+    PBC_flag : str('none', 'hdim_1', 'hdim_2', 'both'), optional
```
freemansw1, line 62: Same comment on optional as above.

freemansw1 (changes requested):
Thanks for doing this, @w-k-jones; a critical improvement to an important piece of code. 

I'm not that familiar with this code, but I've added in some thoughts here in a few places. Probably would be good to get a review in from @kelcyno.

tobac/feature_detection.py
w-k-jones: Looking through the main functions, I think we need to standardise a way of highlighting optional keywords and their default values. I'll update this for now by stating the default at the end of the description

w-k-jones:
Reminder to myself: currently the PBC handling requires all of `h1_min`, `h1_max`, `h2_min` and `h2_max` to be provided even if PBCs are calculated over only one dimension. I should fix this to use default values for the dimension not being used

tobac/merge_split.py
```diff
@@ -7,8 +7,30 @@
     
 """
 
-
-def merge_split_MEST(TRACK, dxy, distance=None, frame_len=5):
+import logging
```
kelcyno, line 10: @w-k-jones you've done a great job cleaning up the code in general, I appreciate the changes/updates in formatting/text.

tobac/merge_split.py
```diff
@@ -32,6 +56,29 @@ def merge_split_MEST(TRACK, dxy, distance=None, frame_len=5):
         Threshold for the maximum number of frames that can separate the end of cell and the start of a related cell.
         Default is five (5) frames.
 
+    cell_number_unassigned: int, optional
```
kelcyno, line 59: I like the addition of being able to assign your own non-cell number.

tobac/merge_split.py
```diff
@@ -54,172 +101,192 @@ def merge_split_MEST(TRACK, dxy, distance=None, frame_len=5):
...
+            )
+            * dxy
+        )
+
+    if PBC_flag in ["hdim_1", "hdim_2", "both"]:
+        # Note that we multiply by dxy to get the distances in spatial coordinates
+        dist_func = build_distance_function(
+            min_h1 * dxy, max_h1 * dxy, min_h2 * dxy, max_h2 * dxy, PBC_flag
+        )
+        cell_start_tree = BallTree(
```
kelcyno, line 163: The BallTree update does not materially change the outputs, but it's faster, great addition.

kelcyno (approved):
I don't have the ability to test this with PBC data, but with radar data and my typical datasets the updates from @w-k-jones works swimmingly. I'm happy to approve as long as Sean's comments are/have been addressed.

kelcyno:
Oh - per our last Tobac meeting we talked about if Will's changes needed to be a separate merge/split method - it does not need to be a separate method.

JuliaKukulies:
> Oh - per our last Tobac meeting we talked about if Will's changes needed to be a separate merge/split method - it does not need to be a separate method.

@w-k-jones @kelcyno I tested this with some model data over CONUS and found that Will's merge/split method results in a significantly lower number (about a 5th) of unique track IDs (basically the variables `feature_parent_track_id` and `cell_parent_track_id`). Not sure if this is a bug but probably not intended behaviour? 

Is the assignment of track IDs here really the same to what Kelcy does in the long loop? 

https://github.com/w-k-jones/tobac/blob/e9ec119b45becbcf0057cfc11114dd71fe3e33ab/tobac/merge_split.py#L205-L213

@w-k-jones I tested this with the MCS criteria for our DYAMOND project and because the number of unique track IDs is so different using the modified merge split function, this results also in a very different number of MCSs if the criteria are applied to the clusters belonging to a track.

w-k-jones:
@JuliaKukulies I believe the change in the results is due to a bug I fixed in the old version: because the minimum spanning tree was applied before the time filter it would result in cell starts/ends which were close together in x/y coords but far apart from time being identified as neighbours, but then subsequently trimmed due to the time gap. This could mean that other cells which were within the time range for merging but further apart in x/y would not be linked. I changed the order in which these operations were applied to avoid this issue, which will result in more cell merges and a smaller number of unique tracks. This wasn’t found originally as the method was designed for short time periods, but became apparent during MCSMIP due to the long time period. Could you test running your data over different length time slices (e.g. 1 hour, 2 hour, 3 hours etc) and see if the proportion of tracks/cells remains roughly the same for the new version but increases (i.e. more tracks and fewer cells per track) with the old version?

JuliaKukulies:
> @JuliaKukulies I believe the change in the results is due to a bug I fixed in the old version: because the minimum spanning tree was applied before the time filter it would result in cell starts/ends which were close together in x/y coords but far apart from time being identified as neighbours, but then subsequently trimmed due to the time gap. This could mean that other cells which were within the time range for merging but further apart in x/y would not be linked. I changed the order in which these operations were applied to avoid this issue, which will result in more cell merges and a smaller number of unique tracks. This wasn’t found originally as the method was designed for short time periods, but became apparent during MCSMIP due to the long time period. Could you test running your data over different length time slices (e.g. 1 hour, 2 hour, 3 hours etc) and see if the proportion of tracks/cells remains roughly the same for the new version but increases (i.e. more tracks and fewer cells per track) with the old version?

@w-k-jones Thanks for this clarification! That makes a lot of sense. I ran tests with different time slices and you are absolutely right. The number of unique tracks and their proportion to cells increases in the old version but the proportion stays about the same in your updated version. The difference in unique tracks is not very high in the first hours and even days, but then becomes quite high when running the merging and splitting module for tracks of a whole month.

JuliaKukulies (approved):
With the clarifications from our last discussion and the additional tests I did, I am approving this now because everything seems to work fine and as expected :) Excellent job, @w-k-jones!

freemansw1 (approved):

Issue #413, opened 2024-03-07
3D feature detection with min_distance fails if dz is not provided 


w-k-jones:
If dz is not given, `filter_min_distance` tries to find the vertical coordinate from the dataframe but as this occurs before coordinate interpolate it fails, causing an error.

Easiest solution is to move this after coordinate interpolation I think

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[22], line 1
----> 1 features_w = tobac.feature_detection_multithreshold(
      2     w_iris, **parameters_features
      3 )

File ~/python/tobac/tobac/utils/decorators.py:168, in xarray_to_iris.<locals>.wrapper(*args, **kwargs)
    165             output = xarray.DataArray.from_iris(output)
    167 else:
--> 168     output = func(*args, **kwargs)
    169 # print(output)
    170 return output

File ~/python/tobac/tobac/feature_detection.py:1375, in feature_detection_multithreshold(field_in, dxy, threshold, min_num, target, position_threshold, sigma_threshold, n_erosion_threshold, n_min_threshold, min_distance, feature_number_start, PBC_flag, vertical_coord, vertical_axis, detect_subset, wavelength_filtering, dz, strict_thresholding, statistic)
   1372     # Loop over DataFrame to remove features that are closer than distance_min to each
   1373     # other:
   1374     if min_distance > 0:
-> 1375         features_thresholds = filter_min_distance(
   1376             features_thresholds,
   1377             dxy=dxy,
   1378             dz=dz,
   1379             min_distance=min_distance,
   1380             z_coordinate_name=vertical_coord,
   1381             target=target,
   1382             PBC_flag=PBC_flag,
   1383             min_h1=0,
   1384             max_h1=hdim1_max,
   1385             min_h2=0,
   1386             max_h2=hdim2_max,
   1387         )
   1388 list_features_timesteps.append(features_thresholds)
   1390 logging.debug(
   1391     "Finished feature detection for " + time_i.strftime("%Y-%m-%d_%H:%M:%S")
   1392 )

File ~/python/tobac/tobac/feature_detection.py:1492, in filter_min_distance(features, dxy, dz, min_distance, x_coordinate_name, y_coordinate_name, z_coordinate_name, target, PBC_flag, min_h1, max_h1, min_h2, max_h2)
   1489 is_3D = "vdim" in features
   1491 if is_3D and dz is None:
[...]

PR #449, opened 2024-09-20
Optional calculation of bulk statistics in feature detection on raw input (before smoothing)

JuliaKukulies:
Addresses #440 and allows for the user to choose whether the bulk statistics in the feature detection are calculated on the unsmoothed or the smoothed input data. 

* [x] Have you followed our guidelines in CONTRIBUTING.md? 
* [x] Have you self-reviewed your code and corrected any misspellings? 
* [x] Have you written documentation that is easy to understand?
* [x] Have you written descriptive commit messages? 
* [x] Have you added NumPy docstrings for newly added functions? 
* [x] Have you formatted your code using black? 
* [x] If you have introduced a new functionality, have you added adequate unit tests?
* [x] Have all tests passed in your local clone? 
* [x] If you have introduced a new functionality, have you added an example notebook?
* [x] Have you kept your pull request small and limited so that it is easy to review? 
* [x] Have the newest changes from this branch been merged?

PR #452, opened 2024-09-25
Fix minimum distance filtering for varying vertical coordinates

### LINKED ISSUE #430: `min_distance` doesn't work with 3D tracking where `dz` not constant
Reported to me by Yishi Hu from OU. When you use `min_distance` in feature detection with a non-constant `dz`, it fails as the coordinate interpolation takes place after `min_distance` is called.  That means `min_distance` doesn't have any vertical information. This should be able to be solved by reversing the order of the two calls. 

w-k-jones:
Resolves #430 and #413

The order of coordinate interpolation and minimum distance filtering has been swapped, so that the z coordinate can be used for finding the spacing of features rather than assuming a fixed vertical grid spacing. The `filter_min_distance` has also been fixed so that the provided z coordinate will be used, rather than always using `dz`. Incidentally, another bug was fixed in `feature_detection_multithreshold_timestep` was fixed, in which statistics were being recalculated at each threshold value, rather than once at the end.

Ideally, I would like to move adding coordinates and distance filtering inside `feature_detection_multithreshold_timestep`, but at the moment we are not passing through the required info to do this. As the feature detection code is in need of refactoring, I'll leave this until after v1.6 to avoid conflicts with the switch to xarray

### DIFF
--- a/tobac/feature_detection.py
+++ b/tobac/feature_detection.py
@@ -627,9 +627,9 @@ def feature_detection_threshold(
                         # find the updated label, and overwrite all of label_ind indices with
                         # updated label
                         labels_2_alt = labels_2[label_z, y_val_alt, x_val_alt]
-                        labels_2[
-                            label_locs_v, label_locs_h1, label_locs_h2
-                        ] = labels_2_alt
+                        labels_2[label_locs_v, label_locs_h1, label_locs_h2] = (
+                            labels_2_alt
+                        )
                         skip_list = np.append(skip_list, label_ind)
                         break
 
@@ -673,9 +673,9 @@ def feature_detection_threshold(
                         # find the updated label, and overwrite all of label_ind indices with
                         # updated label
                         labels_2_alt = labels_2[label_z, y_val_alt, label_x]
-                        labels_2[
-                            label_locs_v, label_locs_h1, label_locs_h2
-                        ] = labels_2_alt
+                        labels_2[label_locs_v, label_locs_h1, label_locs_h2] = (
+                            labels_2_alt
+                        )
                         new_label_ind = labels_2_alt
                         skip_list = np.append(skip_list, label_ind)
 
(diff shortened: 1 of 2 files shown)


w-k-jones:
@freemansw1 are you happy to review this?

freemansw1 (approved):
Thanks, @w-k-jones . Checks out on my end.

tobac/feature_detection.py
```diff
@@ -1121,39 +1121,30 @@ def feature_detection_multithreshold_timestep(
...
 
-        if statistic:
-            # reconstruct the labeled regions based on the regions dict
-            labels = np.zeros(track_data.shape)
-            labels = labels.astype(int)
-            for key in regions_old.keys():
-                labels.ravel()[regions_old[key]] = key
-                # apply function to get statistics based on labeled regions and functions provided by the user
-                # the feature dataframe is updated by appending a column for each metric
-            if statistics_unsmoothed:
```
JuliaKukulies, line 1132: I just noted that the optional calculation on the unsmoothed input data is not merged into your PR yet and since you removed a whole paragraph and placed it somewhere else I bet that these lines would disappear. Maybe the easiest fix is to copy over `if statistics_unsmoothed` and place it within the code paragraph you added after each threshold has been calculated?

JuliaKukulies (approved):
Looking good @w-k-jones! Thanks for also fixing the statistics issue. Just one thing I am wondering with regard to  #449

tobac/feature_detection.py
w-k-jones: Good catch! I'll add that back in

w-k-jones:
I have re-added the unsmoothed statistics functionally which was removed by mistake. The remaining failing tests and the formatting and one of the matrix tests which has the sporadic spectral filtering test bug. @JuliaKukulies & @freemansw1 are you happy for me to merge?

JuliaKukulies:
Yes, ready to merge from my side! Thanks @w-k-jones !

Issue #473, opened 2024-12-13
Minimum requirement for a homepage release


fsenf:
discussed in our dev meeting:

- [ ] meaning image without duplication
- [ ] content-rich landing pages for each entry point from page menu
- [ ] correct API section
- [ ] check that no content from earlier implementation is lost

Please add more requirements

fsenf:
Now, included into #477 .

PR #470, opened 2024-11-15
Feature: Documentation Revamp and Retheme

freemansw1:
After #460, I started to poke around to figure out how to make things look better on our documentation. As it turns out, making the docs look and feel modern was really not all that much work. I also thought v1.6.0 would be an ideal time to produce new, modern-feeling documentation. 

This represents just a first step. I have modified very little text, instead just reorganizing things into categories to fit the theme. Note that this is requested to be merged into a new `dev_documentation_revamp` branch, which comes off `RC_v1.6.0`. I'm putting this in a branch by itself to encourage more collaborative work on this; perhaps we can commit to that repo directly rather than doing it through PRs (until it is time to merge into the RC). 

Given the somewhat unusual path here, I'm requesting relatively *light* reviews. We will continue to make updates before merging into the RC, I am at this point just seeking feedback on whether or not this is broadly fits what we are looking to do with our documentation. 

* [x] Have you followed our guidelines in CONTRIBUTING.md? 
* [x] Have you self-reviewed your code and corrected any misspellings? 
* [x] Have you written documentation that is easy to understand?
* [ ] Have you written descriptive commit messages? 
* [ ] Have you added NumPy docstrings for newly added functions? 
* [ ] Have you formatted your code using black? 
* [ ] If you have introduced a new functionality, have you added adequate unit tests?
* [ ] Have all tests passed in your local clone? 
* [ ] If you have introduced a new functionality, have you added an example notebook?
* [ ] Have you kept your pull request small and limited so that it is easy to review? 
* [ ] Have the newest changes from this branch been merged?

PR #477, opened 2024-12-19
Major Website Update

fsenf:
This is a follow-up of #470  - an imitative to update our website.

Let's try to document further development here, i.e.
- what needs top be done
- and who is working on and implementing updates
as long this is in a draft stage.

Here #473 , we documented minimum requirements for a homepage release. I repeat it to have everything at one place:

- [ ] meaningful images without duplication
- [ ] content-rich landing pages for each entry point from page menu
- [ ] correct API section
- [ ] check passed that no content from earlier implementation is lost


fsenf:
I discussed with @Sven-248 today that he might look into the "Section Navigation" thing.

freemansw1:
Thanks, @fsenf . I've been working on debugging the API page and have it most of the way there. I'll push those changes when it's working.

freemansw1:
I've updated the API pages to work properly now. I've removed some of the documentation pages, especially from anything internal. 

One question I have for the group is whether or not to start hiding functions from the API documentation. Take `tobac.feature_detection.feature_position` as an example. It should only be called internally, I cannot think of a reason for it to be called externally. We can remove it (or any other function) from the docs by adding `:meta private:` to the docstring. I think that would be a good idea for us to do.

fsenf:
@freemansw1 : Great update! API reference looks nice now!

I agree to hide everything that is not primarily used externally. This keeps the API docs clear.

fsenf:
I added our example thumbnails to the entry point of our homepage. There is room for improvement ... What does the group think?

see: https://tobac--477.org.readthedocs.build/en/477/

freemansw1:
> I added our example thumbnails to the entry point of our homepage. There is room for improvement ... What does the group think?
> 
> see: https://tobac--477.org.readthedocs.build/en/477/

Thanks, @fsenf . That is a significant improvement. I think, broadly, we should find a way to have the homepage take up more of the page, and have less whitespace. Perhaps that means switching homepage templates, perhaps that just means expanding the content some. Right now, I'm working on figuring out hiding private API functions (`:meta private:` did not work by itself).

freemansw1:
Per our discussion last meeting, I've moved this to post-v1.6. Per our usual policy on documentation, we can release this without a version increment, but I think an email announcement would be nice.

freemansw1:
Switched merge base to `RC_v1.6.x` to solve testing issues driven by #531

fsenf:
Based on @JuliaKukulies 's comment, I checked and updated the sphinx-based rendering docu. All is correct now.

freemansw1:
Thanks, @JuliaKukulies . I've made the revisions you requested. Thanks for the detailed look through!

> * When you access the API references through the "API" tab in the top menu, it works, but the methods for the `tobac.feature_detection` module is incomplete. Not sure why this is only the case for the feature detection and not for the other modules
> *

I'm not sure what you mean here. If it is what I think it is, I diliberately hid most of the functions in feature detection, as they really shouldn't be used by end-users. Let me know if you want more of the functions visible.

Issue #478, opened 2025-01-10
a good 3d example is missing


fsenf:
Not sure if I overlooked it, but a cool 3d example would be very helpful.

kelcyno:
I can take this on during the 2025 Tobathon

fsenf:
Great! I know that @Sven-248 started to work on this topic. Not sure how far he got, but a good chance to join forces.

Sven-248:
Yes i've started working on this. Feel free to check out the current progress here:
https://github.com/Sven-248/tobac/tree/3d-blob-tracking-example

I plan to add segementation and statistical analysis tomorrow.

kelcyno:
@Sven-248 That's wonderful! I have some data I'm pre-processing for a non-idealized example. I'm hoping it builds on your work, your example is excellent.

mathilde-ritman:
I'd also be happy to help on this for the Tobathon today, if you're in need of a pair-coder Kelcy. Otherwise, I could work to contribute a 3d animation for the webpage enhancements. Talk soon.

Sven-248:
@kelcyno I'm also interested in helping with this. I can contribute to building on the example as well. Let me know how i can assist!

Issue #37, opened 2020-04-17
Calculation of velocity components-vectors


mheikenfeld:
Currently, the analysis function calculating the velocity of the tracked cells only provides the absolute value of the velocity. Velocity components in x-y/lat-lon directions should be relatively easy to implement...

PR #500, opened 2025-05-14
Add new 3D blob tracking example

Sven-248:
This PR introduces a new example showcasing 3D blob tracking (https://github.com/tobac-project/tobac/issues/478). In addition, the following updates have been made:

* New example added: A new example for generating and tracking 3D blobs has been added to the example folder

* Environment update: The environment-examples.yml file has been updated to include any necessary dependencies related to the new example

* Test function: A new test function has been added to tobac/testing.py to create a 3D blob

* Doc: The new 3d example has been added to the gallery with an appropriate thumbnail

* [ ] Have you kept your pull request small and limited so that it is easy to review? 


Sven-248:
I recommend handling issue #37 (Calculation of velocity components-vectors) in a separate PR, as it seems to affect multiple example scripts beyond the one modified here.

JuliaKukulies (approved):
Thank you so much for this nice and comprehensive example @Sven-248 !! I think it definitely is a great addition to the basic examples and really helps to understand some of the tobac capabilities in more detail based on an easy idealized case. 

I am happy for this to get merged, but I do have two small comments to the notebook that you may or may not address before merging: 

- Would it be possible to replace Section 5 with an example of `tobac.get_bulk_statistics_from_mask()` ? It is also OK to leave the velocity example, but the text directs the reader very much to our analysis module which currently has many functions that do not work anymore and we wanted to get away from this a little bit. The bulk statistics capability should allow for a variety of analyses as well since it makes it easy to derive some statistics of the values of each segmented feature. 
- I think the following section could be shortened down a little since we mention this at various locations and I think "on further extensions" could be more clear, e.g. "two different ways/workflows/orders of using the tobac modules": 

```
Just a small thought here:

We often introduce the different steps as 1.feature detection, 2.segmentation and 3. tracking, so having the segmenation step actually before the trajectory linking. The order actually does not matter since the current segmentation approach is based on the feature detection output only and can be done before or after the linking.

On further extensions:

Also, one could actually use the output of the segmentation (segments Dataset in the example below) as input for the tracking with the advantage that information on the volume (ncells) is added in the dataframe.

One could also use the output of the tracking in the segmentation (trajectories Dataset from above) with the advantage that mask will contain only the features that are also linked to trajectories.

These possibilities exist and could be utlized by you ...

```

PR #512, opened 2025-07-12
Add utilities to convert feature masks to cell masks and vice versa, and add optional cell mask output from segmentation

### LINKED ISSUE #508: Allow users to have segmentation mask the cell number rather than the feature number
Many users (including me) are now running segmentation after tracking. We should allow users to request that segmentation output correspond to the `cell` value rather than the `feature` value.

w-k-jones:
Adds features requested in #508, along with postprocessing utilities to perform the conversion in both directions. Note that in the presence of stub cells, this operation cannot be performed perfectly.

Some additional features not currently included are conversion to track masks (although the reverse is not possible) and a function for simultaneously filtering features from both a feature dataframe and the corresponding feature mask.


JuliaKukulies (approved):
Very useful addition, @w-k-jones! I have tested this with our example notebooks and it works without problems. The design looks also good to me! 

I am just wondering if there was a reason why we are not adding the capability to have either cell or track numbers in the mask (and limit the conversion from features to track to a one-direction conversion only). Same is true for filtered features. With the function that you have for converting features to cell numbers, it seems like an easy add to instead of the cell numbers use tracks or filtered features. 

Also, what do you think about modifying one of our example notebooks in a way that we apply this new function. I think as of now, most notebooks so the tracking first and then the segmentation, but it would be fairly easy to change one example to the tracking first and then show how the returned mask in the segmentation contains the cell numbers. I am happy to help with such an example!

w-k-jones:
I have added functions to convert feature/cell masks to tracks, but no operation in the other direction is possible (individual features get merged together and become inseparable). I have also added an `inplace` keyword to perform this operation without copying the input mask to save memory (e.g. when performed during segmentation).

Ready for re-review when tests have passed @JuliaKukulies @freemansw1

tobac/tests/test_utils_mask.py
```diff
@@ -0,0 +1,1382 @@
...
+
+    test_features = pd.DataFrame(
+        {
+            "feature": [1, 2, 3],
+            "frame": [0, 1, 2],
+            "time": pd.date_range(
+                datetime(2000, 1, 1, 0), datetime(2000, 1, 1, 2), periods=3
+            ),
+            "cell": [1, 1, 1],
+            "track": [1, 1, 1],
```
JuliaKukulies: The functions `convert_feature_mask_to_track` and `convert_cell_mask_to_track` require the format of the original `Features` pandas dataframe with an added column `track` it seems like. As in one of your example notebooks (MCS tracking in ICON simulations), the suggested workflow is: 

```
merge_splits = merge_splits.merge_split_MEST(Track, dxy = dxy)
Track['track'] = merge_splits.feature_parent_track_id
```

which I think is good but it needs to be documented somewhere. From a user perspective, it is not clear whether to input the output of `merge_splits.merge_split_MEST` or do that extra step. I am making this comment here, because maybe we should include this in the test as well; to ensure the output format of merging and splitting can be used with this function.

JuliaKukulies (changes requested):
Thanks for the additions @w-k-jones, great job and so useful!! See my specific comments below. The main thing is that we need to either write clear documentation which extra steps need to be done to add the `track` column to the dataframe of tracked features, or the function needs to be changed so it works with the original output from the merge split function. I think the first option would be easier, but maybe the second one cleaner - what do you think?

freemansw1:
Given the complexities this has opened, I've updated the milestone to 1.6.3. I'll try to get 1.6.2 released in the next couple of days.

w-k-jones:
@JuliaKukulies I have reverted the PR remove the conversion to track masks pending updates to how merge/split output is handled. Would you be happy to re-review? Thanks! :)

PR #542, opened 2025-12-14
Add Latitude/Longitude based tracking

freemansw1:
New function `tobac.tracking.linking_trackpy_latlon`, which uses latitude and longitude (instead of `hdim_1`, `hdim_2`, and `dxy`) to calculate distances. The result runs about 5-10 times faster than an equivalent approach with periodic boundaries for global data. 

This function also has the advantage of setting us up well for unstructured grid data. 


freemansw1:
Requesting @JuliaKukulies for review.

Anticipate releasing this as part of the next v1.6.x; I don't think this merits holding for v1.7.0.

tobac/tracking.py
```diff
@@ -492,6 +495,410 @@ def linking_trackpy(
     return trajectories_final
 
 
+def linking_trackpy_latlon(
+    features: pd.DataFrame,
+    dt: float,
+    v_max: float = None,
+    d_max: float = None,
+    latitude_name: str = "",
+    longitude_name: str = "",
```
w-k-jones: Default value should be `None`, rather than an empty string (particularly as lines 740-743 assigns `None` to both these variables if they are an empty string)

tobac/tracking.py
```diff
@@ -492,6 +495,410 @@ def linking_trackpy(
...
+        )
+
+    # Reset trackpy parameters to previously set values
+    if subnetwork_size is not None:
+        if adaptive_step is None and adaptive_stop is None:
+            tp.linking.Linker.MAX_SUB_NET_SIZE = size_cache
+        else:
+            tp.linking.Linker.MAX_SUB_NET_SIZE_ADAPTIVE = size_cache
+
+    filtered_trajectories = _filter_trajectories(
```
w-k-jones, line 771: We should also update the x/y tracking function to use this to avoid duplicate code

w-k-jones (changes requested):
Great addition! Just two minor changes requested, but otherwise the code and examples are looking great.

w-k-jones:
One other minor change I just noticed: the `tracking._filter_trajectories` function is showing up in the API reference in the docs, we should probably hide this

w-k-jones:
As a separate issue from this PR, we should allow the default linking algorithm to use the feature coordinates, rather than hdim_1, hdim_2 etc for tracking (as part of the desire to remove `dxy` from _tobac_...)

freemansw1:
Thanks for your review, @w-k-jones . I've made the changes that you requested. 

> As a separate issue from this PR, we should allow the default linking algorithm to use the feature coordinates, rather than hdim_1, hdim_2 etc for tracking (as part of the desire to remove `dxy` from _tobac_...)

Yes, agreed, but that sort of auto-detection is nontrivial (and we'd have to communicate that well to users).

tobac/tracking.py
```diff
@@ -427,10 +431,387 @@ def linking_trackpy(
...
+            tp.linking.Linker.MAX_SUB_NET_SIZE = subnetwork_size
+        else:
+            size_cache = tp.linking.Linker.MAX_SUB_NET_SIZE_ADAPTIVE
+            tp.linking.Linker.MAX_SUB_NET_SIZE_ADAPTIVE = subnetwork_size
+
+    # find latitude and longitude columns
+    if latitude_name == "":
+        latitude_name = None
+    if longitude_name == "":
+        longitude_name = None
```
w-k-jones: I think that the checks for empty string are now unnecessary, and should be removed.

w-k-jones:
> Thanks for your review, @w-k-jones . I've made the changes that you requested.
> 

Great, thanks, now I think the only leftover change I can see is removing the unnecessary checks for empty string inputs

> > As a separate issue from this PR, we should allow the default linking algorithm to use the feature coordinates, rather than hdim_1, hdim_2 etc for tracking (as part of the desire to remove `dxy` from _tobac_...)
> 
> Yes, agreed, but that sort of auto-detection is nontrivial (and we'd have to communicate that well to users).

As a first step, we could add `x_name` and `y_name` keywords to `linking_trackpy`, and fall back to using hdim1, hdim2 and dxy if not provided. Long term, this is something that using xarray datasets rather than pandas pataframes for storing feature/track info in, as we could store the core coordinate names found during feature detection as metadata which can be used by tracking and other processes.

w-k-jones (approved):
Happy for this to be merged once all the checks have passed

PR #550, opened 2026-01-08
Return empty DataFrame when no features are detected

### LINKED ISSUE #510: Return a properly formatted DataFrame when no features are detected

Right now, if you detect no features, `None` is returned. Instead, we should return an empty (but properly formatted) dataframe.

Sven-248:
Fixes #510 .

When no features are detected, return an empty but properly formatted DataFrame instead of None.
Adds tests for 2D and 3D feature detection and internal utilities.

### DIFF
--- a/tobac/feature_detection.py
+++ b/tobac/feature_detection.py
@@ -1555,9 +1555,11 @@ def feature_detection_multithreshold(
                 label_fields[i].data[~wh_all_labels] = 0
 
     else:
-        features = None
-        label_fields = None
-        logging.debug("No features detected")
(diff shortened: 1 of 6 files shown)


freemansw1:
> I'll get this reviewed later today or tomorrow before the meeting. It would be good for us to extend the combine_feature_dataframes function to recognize empty dataframes and increment e.g., frame properly as well.

I've realized that this doesn't work very easily because e.g., only one empty dataframe would be generated, which makes sense, but isn't ideal for combine_feature_dataframes. We won't be able to do this piece (easily), but it's worth the conversation about making the metadata better.

freemansw1 (approved):
I think this is a useful and important change. It's well implemented and ready to merge.

PR #551, opened 2026-01-19
Add ability to detect "families"

freemansw1:
I've talked about this before; this is (finally) the code for family detection (family tracking still to come). This code links features together in space. 

For the reviewers (I'm leaning toward two reviewers given that this is a new concept for tobac), please check the following:

- [ ] Documentation makes sense
- [ ] Example works and makes sense
- [ ] This works for your data 

Note: this is targeted at a new branch, `RC_v1.7.0`. 

* [x] Have you followed our guidelines in CONTRIBUTING.md? 
* [x] Have you self-reviewed your code and corrected any misspellings? 
* [x] Have you written documentation that is easy to understand?
* [x] Have you written descriptive commit messages? 
* [x] Have you added NumPy docstrings for newly added functions? 
* [x] Have you formatted your code using black? 
* [x] If you have introduced a new functionality, have you added adequate unit tests?
* [x] Have all tests passed in your local clone? 
* [x] If you have introduced a new functionality, have you added an example notebook?
* [x] Have you kept your pull request small and limited so that it is easy to review? 
* [x] Have the newest changes from this branch been merged? 

@w-k-jones and/or @JuliaKukulies are you up for reviewing? I know this is for v1.7.0, but I'd like to keep this moving so I can implement family tracking soon.

PR #556, opened 2026-01-30
Implementation of Geometric Subsegmentation into Feature Detection

### LINKED ISSUE #565: Bug in Segmentation Testing

With my subsegmentation PR, I encountered CI problems (see https://github.com/tobac-project/tobac/actions/runs/22542485302/job/65299673049) that I could trace back to segmentation testing.

After removing this line:
https://github.com/tobac-project/tobac/blob/fcb7cd307b9adf241f2b7bc2ef78cb306eb69731/tobac/tests/segmentation_tests/test_segmentation_time_pad.py#L77
all testing seem to work fine.

@freemansw1 : I believe you contributed this test. Could you confirm that bug and its re-solution?

fsenf:
This has been on my list for years now. I am really interested to get your feedback on any aspect. 

**Content**:
Geometric subsegmentation is an iterative method that works with a single threshold and breaks down the field into smaller parts. Its implementation is scale invariant, i.e. feature that appear similar are treated the same way independent of their size (and of course discretization effects destroy this property for very small objects). The method has been applied for analyzing tropical anvil clouds here:

Senf, F., D. Klocke, and M. Brueck, 2018: Size-Resolved Evaluation of Simulated Deep Tropical Convection. Mon. Wea. Rev., 146, 2161–2182, https://doi.org/10.1175/MWR-D-17-0378.1. 

**Implementation**:
- I made a new submodule, here `tobac/segmentation/geometric_subsegmentation.py` and just copied over my still existing functions. Likely, there is functionality the is duplicate to existing one --> How to deal with this?
- I am a big fan of curvature flow filters. I kept this in here and this adds a dependency to https://simpleitk.org/about.html

**Testing**:
- unit tests are missing still
- an example notebook shows the principle. This case is indeed not to different to multi-threshold feature detection. Maybe I need to think about a show case that makes the added values more clear ...
- If you like to run it for your one, please care that your env includes `simpleitk`, e.g. via `mamba install simpleitk`

This is really an early drafting stage - also about the question how different feature detection pathways might look like. Please be critical and actively discuss.


fsenf:
@freemansw1 , @w-k-jones , @JuliaKukulies , @kelcyno : ping for you in case you are interested.

fsenf:
@freemansw1 , @w-k-jones , @JuliaKukulies , @kelcyno : ping you again. Do you really have no opinion on this development idea?

w-k-jones:
> @freemansw1 , @w-k-jones , @JuliaKukulies , @kelcyno : ping you again. Do you really have no opinion on this development idea?

I've only had time to have a brief look at this, but it appears very nice! I think it is very complementary to Sean's family detection in #551 , it might be a nice addition to add a keyword to allow preserving the original feature labels as families after subsegmentation. Let me know if you would like a more in depth review on this as it is currently, and I can put some time aside for it later this week :)

Issue #481, opened 2025-03-03
Segmenting failure with diagonal pixels


nickmesa:
It seems as though tobac struggles to fully identify and segment a feature if there are diagonal pixels adjacent to the "main" part of the feature. This leaves data unaccounted for, as it is not attached to this "main" feature, nor is it identified as a standalone feature itself. Are there any workarounds to this? Thanks!

freemansw1:
Can you provide a figure/example of what you mean here? Is your data on a regular square grid, or is the grid diagonal?

freemansw1:
Okay, this is what I get for trying to do this in the middle of the night. I think a figure would still help here, but I now understand the issue. The short answer is that no, we don't yet support that. 

It would not be that challenging to add (and in fact we should probably add it to feature detection, too). We could allow a user-set parameter in segmentation that feeds into `connectivity` in `skimage.segmentation.watershed`. 

I would be supportive of a pull request that added a capability like that, although I don't have bandwidth to add it myself at the moment. @nickmesa, I'd be happy to help you add it in via a pull request, though.

w-k-jones:
If I remember correctly, the labelling during feature detection does include diagonal connectivity (by default it uses connectivity 2 vs 1 for segmentation). This is another case of discrepancies between thresholding in feature detection which we should address. I agree with making connecitivity a customisable parameter, but I think the default should be connectivity=1 (i.e. no diagonal connections)

freemansw1:
> If I remember correctly, the labelling during feature detection does include diagonal connectivity (by default it uses connectivity 2 vs 1 for segmentation). This is another case of discrepancies between thresholding in feature detection which we should address. I agree with making connecitivity a customisable parameter, but I think the default should be connectivity=1 (i.e. no diagonal connections)

Yes, that seems to be right (although the documentation is confusing). I am always hesitant to change defaults on science things without warning users. We will need to give ample warning, etc. Maybe the easiest thing to do is to expose this parameter (in segmentation and feature detection) to the user, keep the defaults the same for now, and start warning people to specify a parameter.

This could be a good tobathon topic, too, and something good to release post v1.6.0.

nickmesa:
<img width="570" alt="Image" src="https://github.com/user-attachments/assets/9f02c4c0-4ca0-4af0-8ef4-f1533aab7c40" />

[...]

PR #562, opened 2026-02-05
Replace PBC handling in feature detection with label function from dask-image

w-k-jones:
The `dask_image.ndmeasure.label` function replicates the `scipy.ndimage.label` function, but adds a `wrap_axes` keyword that can be used to perform label detection across periodic boundaries (it requires identical logic to labelling regions across tiles). Replacing the current feature detection PBC code with this both simplifies our codebase, and starts to integrate more dask features into the tobac pipeline so that eventually it can be fully dask enabled. I have set up the `label` function to replicate the handling of connectivity indentically to the scikit-image function it is replacing, meaning that full connectivity is used rather than square connectivity. However, in future I think we should make this an option and change the default to square (1) connectivity to match segmentation (see #481)

Currently in draft as I am not sure whether to introduce this on its own or as part of a larger refactor of feature detection

* [ ] Have you followed our guidelines in CONTRIBUTING.md? 
* [ ] Have you self-reviewed your code and corrected any misspellings? 
* [ ] Have you written descriptive commit messages? 
* [ ] Have all tests passed in your local clone? 
* [ ] Have you kept your pull request small and limited so that it is easy to review? 

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -124,15 +124,15 @@ def setup(app):
 
 # This should include all modules used in tobac. These are dummy imports,
 # but should include both required and optional dependencies.
+# NOTE: this should use the name of the package as it is imported into python, not the name on conda/pypi
 autodoc_mock_imports = [
     #    "numpy",
     "scipy",
-    "scikit-image",
     "pandas",
     "pytables",
     "matplotlib",
     "iris",
-    "cf-units",
+    "cf_units",
     "xarray",
     "cartopy",
     "trackpy",
@@ -141,7 +141,6 @@ autodoc_mock_imports = [
     "sklearn",
     "cftime",
     "dask",
-    "dask-image",
     "dask_image", 
 ]
 
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -140,6 +140,8 @@ autodoc_mock_imports = [
     "skimage",
     "sklearn",
     "cftime",
+    "dask",
+    "dask-image",
 ]
 
 autodoc_default_options = {
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -142,6 +142,7 @@ autodoc_mock_imports = [
     "cftime",
     "dask",
     "dask-image",
+    "dask_image", 
 ]
 
 autodoc_default_options = {
--- a/environment-ci.yml
+++ b/environment-ci.yml
@@ -15,6 +15,8 @@ dependencies:
   - trackpy
   - pytest
   - typing_extensions
+  - dask
+  - dask-image
   - black
   - coverage
   - pytest-cov
--- a/environment-examples.yml
+++ b/environment-examples.yml
@@ -14,6 +14,8 @@ dependencies:
   - trackpy>=0.6.1
   - pytest
   - typing_extensions
+  - dask
+  - dask-image
   - black
   - jupyter
   - notebook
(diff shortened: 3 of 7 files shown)


w-k-jones:
Testing with the low cloud tracking notebook shows a slight slow down in performance:

Previous: `3.43 s ± 27.5 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)`
New: `3.61 s ± 36.8 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)`

I will test on a larger example, but I think this is a reasonable trade off for simpler code

w-k-jones:
On further investigation with larger datasets, it seems like there is a fairly large overhead to using the dask image label function for moderately large feature detection jobs (where feature detection takes ~10 seconds to 1 minute), but performance benefits for larger tasks (e.g. multiple minutes). I will investigate a bit more to see if time steps vs spatial array size has different results. Note that this is using the default dask settings, so not increasing the number of workers etc. or chunking the data.

w-k-jones:
Some test results:

MCSMIP Obs (geo-ir), 1200x3600 domain, PBC hdim_2: [on HPC]
24 time steps: previous: `26.92 s`, new: `48.26 s`
240 time steps: previous: `04:41.61`, new: `04:12.25`
672 time steps: previous: `0:11:17.83`, new: `09:29.44`

EUREC4A LES, 1524x1524 domain, PBC both: [on laptop]
13 time steps: previous: `1.04 s ± 6.93 ms per loop`, new: `1.63 s ± 20.5 ms per loop`
49 time steps: previous: `3.66 s ± 27.1 ms per loop`, new: `5.84 s ± 71.5 ms per loop`
720 time steps: previous: `41.8 s ± 528 ms per loop`, new: `1min 22s ± 741 ms per loop`

What's puzzling is that the relative performance is changing with the number of timesteps, even though the labelling is working independently per timestep. Dask continues to be an enigma to me, do you have any ideas on this @freemansw1 ?

freemansw1:
@w-k-jones I'm certainly not a dask expert (and I'm looking at this relatively quickly), but I'm guessing it has to do with how the dask graph is constructed/how big the dask graph gets. Even if it's operating independently per timestep, unless `compute` is being run after every timestep, it only (in theory) does it at the end. I've also found that sometimes if you *do* run `compute` every timestep, it recomputes old results. I'm not sure why it does that, though. 

This is very exciting, though. I wonder how performance scales by number of workers.

w-k-jones:
> @w-k-jones I'm certainly not a dask expert (and I'm looking at this relatively quickly), but I'm guessing it has to do with how the dask graph is constructed/how big the dask graph gets. Even if it's operating independently per timestep, unless `compute` is being run after every timestep, it only (in theory) does it at the end. I've also found that sometimes if you _do_ run `compute` every timestep, it recomputes old results. I'm not sure why it does that, though.
> 
> This is very exciting, though. I wonder how performance scales by number of workers.

Currently I have it running compute every timestep immediately for the labelling, to avoid having change any of the surrounding code, and given the array isn't chunked it will likely run slower than the scipy/scikit-image labelling just due to overhead. 

From a quick inspection dask-image should have substitutes for most of the image functions we need for feature detection. I'll see if I can get lazy execution of the per timestep feature detection working in a simple manner and see if running with an actual dask client and changing the number of workers affects things.

w-k-jones:
I think that the best way forward here is to keep both a "small data" approach based on the current method for performance on smaller problems, and a "big data" alternative using dask-image that is used if a dask array is passed to feature detection.

PR #575, opened 2026-06-16
Add method to link tracking performed on two different fields (i.e. "multivariate tracking")

mathilde-ritman:
As mentioned at last month's dev meeting and in [this discussion](https://github.com/tobac-project/tobac/discussions/573), I've now written up my work linking tracks of two different fields (e.g., an anvil with its convective cores).

I've attempted to meet all the criteria below. That said, this is my first time contributing to a project on git, so do let me know if I've missed anything obvious! Also included are new unit tests and an example notebook.

We had discussed aiming this at branch RC_v1.7.0.


tobac/tests/test_multivariate.py
```diff
@@ -0,0 +1,290 @@
...
+        "index": np.array([1, 2], dtype=np.int64),
+    }
+
+    out = tobac.multivariate.find_mappings_for_chunk(chunk, nan_val=99)
+
+    assert (out == np.array([3, 2], dtype=np.int64)).all()
+
+
+def test_drop_repeated_mappings():
+    """Assure that drop_repeated_mappings maps each child label uniquely and to the minimum parent label, recording the laternatives"""
```
w-k-jones, line 22: Typo:
```suggestion
    """Assure that drop_repeated_mappings maps each child label uniquely and to the minimum parent label, recording the alternatives"""
```

tobac/multivariate.py
```diff
@@ -0,0 +1,254 @@
...
+
+    a_mappings = {}  # i: min(j); for mapping i->min(j)
+    j_values = {}  # i: all js;  for mapping j->min(j)
+    for i, j in zip(list_A, list_B):
+        if i in a_mappings:
+            a_mappings[i] = min(a_mappings[i], j)  # parent to min j
+            j_values[i] += [j]  # Collect all j values for i
+        else:
+            a_mappings[i] = j  # First occurrence of i
+            j_values[i] = [j]  # Start collecting j values for i
```
w-k-jones, line 177: Suggest using defaultdict here instead:
```suggestion
    a_mappings = defaultdict(lambda: np.inf)  # i: min(j); for mapping i->min(j)
    j_values = defaultdict(list)  # i: all js;  for mapping j->min(j)
    for i, j in zip(list_A, list_B):
        a_mappings[i] = min(a_mappings[i], j)  # parent to min j
        j_values[i] += [j]  # Collect all j values for i
```

tobac/multivariate.py
```diff
@@ -0,0 +1,254 @@
...
+    # collect map values to format j: min(j)
+    b_mappings = {}
+    for i, js in j_values.items():
+        min_j = a_mappings[i]
+        for j in js:
+            b_mappings[j] = min_j
```
w-k-jones, line 183: There is an issue with this logic: if a parent is linked to multiple children, but one of those children is linked to a smaller number parent, the parent will get relabelled even if it is the parent of another child. e.g. if parent `1` is linked with child `3` and parent `2` is linked to children `3` and `4`, then the resulting parent map will be `{1:1, 2:1}`. We should add a check that if `j` is in `a_mappings.values()` then `b_mappings[j] = j`.

tobac/multivariate.py
```diff
@@ -0,0 +1,254 @@
...
+            child_labels.max(initial=0),
+            parent_labels.max(initial=0),
+        )
+        + 1000
+    )
+
+    logging.info("Start calculating the coincident label mappings")
+
+    child_mask = child_mask.where(child_mask > 0).fillna(NAN_VAL).astype(int)
+    parent_mask = parent_mask.where(parent_mask > 0).fillna(NAN_VAL).astype(int)
```
w-k-jones, line 72: Suggestion: use `other` keyword for `where` rather than `fillna`
```suggestion
    child_mask = child_mask.where(child_mask > 0, other=NAN_VAL).astype(int)
    parent_mask = parent_mask.where(parent_mask > 0, other=NAN_VAL).astype(int)
```

tobac/multivariate.py
```diff
@@ -0,0 +1,254 @@
...
+
+    child_name = child_mask.name  # expect "cell" / "updraft"
+    parent_name = parent_mask.name
+
+    child_tracks[child_name] = np.where(
+        child_tracks[child_name] > 0, child_tracks[child_name], 0
+    )
+    parent_tracks[parent_name] = np.where(
+        parent_tracks[parent_name] > 0, parent_tracks[parent_name], 0
+    )
```
w-k-jones, line 56: Suggest removing this as it is not necessary could cause issues with data mutation due to inplace modification of dataframes.

tobac/multivariate.py
```diff
@@ -0,0 +1,254 @@
...
+    # list all labels to update
+    child_labels = np.unique(child_tracks[child_name].values)
+    parent_labels = np.unique(parent_tracks[parent_name].values)
+    NAN_VAL = (
+        max(
+            child_labels.max(initial=0),
+            parent_labels.max(initial=0),
+        )
+        + 1000
+    )
```
w-k-jones, line 67: Suggest removing unique calculation as the `child_labels` and `parent_labels` variables are only used to calculate the max label values
```suggestion
    NAN_VAL = (
        max(
            child_tracks[child_name].values.max(initial=0),
            parent_tracks[parent_name].values.max(initial=0),
        )
        + 1000
    )
```

tobac/multivariate.py
```diff
@@ -0,0 +1,254 @@
...
+            k: {
+                **label_arrays,
+                "index": k,
+            }
+            for k in unique_child_values
+        }
+        child_mappings = joblib.Parallel(n_jobs=-1, prefer="threads")(
+            joblib.delayed(find_mappings_for_chunk)(label_arrays[k], NAN_VAL)
+            for k in unique_child_values
+        )
```
w-k-jones, line 113: Would remove the parallel calls internally or make it optional

tobac/multivariate.py
```diff
@@ -0,0 +1,254 @@
...
+        map_as_array = np.full(parent_tracks[parent_name].max() + 1, 0, dtype=np.int64)
+        for k, v in map_as_dict.items():
+            map_as_array[k] = v
+        return map_as_array[vals.astype(int)]
+
+    parent_mapped = xr.DataArray(
+        parent_darray.map_blocks(_map_parent, dtype=np.int64),
+        dims=parent_mask.dims,
+        coords=parent_mask.coords,
+    ).rename("multivariate")
```
w-k-jones, line 252: Currently this always performs merging of parents in the output, I think that it would be better if this was made optional with the default for parents to not be relabelled. Potentially there should also be an option to drop children/parents that are not linked with another.

tobac/multivariate.py
```diff
@@ -0,0 +1,254 @@
...
+    parent_name = parent_mask.name
+
+    child_darray = dask.array.from_array(child_mask.fillna(0))
+    parent_darray = dask.array.from_array(parent_mask.fillna(0))
+
+    def _map_child(vals):
+        map_as_dict = child_tracks.set_index(child_name)["multivariate"].to_dict()
+        map_as_array = np.full(child_tracks[child_name].max() + 1, 0, dtype=np.int64)
+        for k, v in map_as_dict.items():
+            map_as_array[k] = v
```
w-k-jones, line 232: Suggest direct assignment rather than via dict
```suggestion
        map_as_array = np.full(child_tracks[child_name].max() + 1, 0, dtype=np.int64)
        map_as_array[child_tracks[child_name].values] = child_tracks["multivariate"].values
```

tobac/multivariate.py
```diff
@@ -0,0 +1,254 @@
...
+        child_darray.map_blocks(_map_child, dtype=np.int64),
+        dims=child_mask.dims,
+        coords=child_mask.coords,
+    ).rename("multivariate")
+
+    def _map_parent(vals):
+        map_as_dict = parent_tracks.set_index(parent_name)["multivariate"].to_dict()
+        map_as_array = np.full(parent_tracks[parent_name].max() + 1, 0, dtype=np.int64)
+        for k, v in map_as_dict.items():
+            map_as_array[k] = v
```
w-k-jones, line 245: Suggest direct assignment rather than via dict
```suggestion
        map_as_array = np.full(parent_tracks[parent_name].max() + 1, 0, dtype=np.int64)
        map_as_array[parent_tracks[parent_name].values] = parent_tracks["multivariate"].values
```

w-k-jones (changes requested):
Hi @mathilde-ritman , thanks for opened this PR, it's a really cool feature! Sorry for the delay getting around to reviewing it. I've made some suggestions for code changes, and found a potential issue with part of the logic in finding the best match for children/parent pairs when a parent has multiple children which themselves have multiple parents.

Just to check, this is my understand of the two main functions, `get_multivariate_label_maps` and `apply_multivariate_label_maps`:

`get_multivariate_label_maps` finds the overlaps between child and parent label arrays at each time step. For each child label the smallest value parent label is selected, and this value is assign to the `multivariate` column in `child_tracks`. For every parent label that is not the smallest cell number overlapping with that child, the `multivariate` column is assigned the value of the smaller parent label.

`apply_multivariate_label_maps` then maps the mask arrays for each of child and parent labels to the value given by the `multivariate` column.

Some questions about this:

1. Can these functions be applied to the results of feature detection/segmentation without tracking? i.e. linking child and parent features rather than cells?
2. Would it be good to implement alternative methods for choosing the best child/parent match than the current approach of choosing the smallest parent label. E.g. choosing the label with the largest overlap area? This could be a later addition
3. At present `apply_multivariate_label_maps` performs merging on the parent objects, is the desired default behaviour? It might be good to make it not relabel parents by default to avoid causing unexpected merging.
4. Handling the logic for parents that have multiple children which themselves have multiple parents. This currently could cause issue. The simplest solution is to avoid assign a parent of a child any other value than itself in the `multivariate` column. It may be desirable in future to be able to fully merge these systems using an approach such as the connected components search used in `merge_split_MEST`

